Issue #5, opened 2015-03-18
Check for fortran is not needed.


anthony-arnold:
As far as I can tell, there's no actual fortran code in this project, and the check for gfortran is there because of the requirement for LAPACK, BLAS and Armadillo. 

I'm currently working on a buildroot configuration to install gnss-sdr on embedded devices. Fortran has been deprecated in buildroot, and the LAPACK and BLAS packages have been replaced by their C equivalents. Additionally, Armadilo can be built without a fortran compiler.

I propose that simply checking for LAPACK and BLAS is enough, and there's no need to check for fortran. This would allow configurations where non-fortran versions of LAPACK and BLAS are used.

carlesfernandez:
Fixed. Now gfortran is not required.

PR #6, opened 2015-03-20
Cross-compilation fixes

anthony-arnold:
These are a few changes which were needed in order to get cross-compilation working a little easier. In particular, when trying to compile in a Buildroot environment.

Here are some justifications for all of the changes:

## 1. Reverted In-tree Build to a Warning
Buildroot likes to do its builds in-tree. It's not really worth the effort to work around it, so I think a simple warning would suffice.

## 2. CMake Arguments for the VOLK module
The C compiler argument was missing, which caused CMake to pick the system default; this breaks if you want to use something different. I also added a toolchain file argument when one is being used. The VOLK module should now build under the same environment as GNSS-SDR.

## 3. Not Failing When gfortran is Not Found
As per issue #5, gfortran isn't actually essential. If gfortran is not available, Armadillo will still build just fine. A warning will still be printed, but config/build will still go ahead.

The main issue here is that Fortran support is deprecated in Buildroot.

## 4. New Compilation Tests in TestForARM and TestForSSE.

When using certain compilers, including ccache, the arguments `-dM` and `-c` are not directly supported (i.e. ccache will balk and refuse to continue.) The Buildroot toolchain file specifices `${CMAKE_CXX_COMPILER_ARG1}` which is the actual path to the cross-compiler and the first argument to ccache. The test compilation now works as expected, and doesn't intrude on standard builds either.

## 5. Setting C++ Compiler Flags for VOLK

The new math functions like `std::exp2` and `std::log`, which were introduced in C++11, are used in Armadillo header files. Added `-std=c++11` to allow compilation to continue.

## 6. Patch for volk_gnsssdr_malloc

VOLK's use of `posix_memalign` is broken on some ARM systems. This is due to the alignment on these systems being 1, which is not a multiple of `sizeof(void*)`. A near-identical patch has been applied to the upstream out-of-tree VOLK for gnuradio.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@
 # Project setup
 ########################################################################
 if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
-    message(FATAL_ERROR "Prevented in-tree build. This is bad practice. Try 'cd build && cmake ../' ")
+    message(WARNING "In-tree build is bad practice. Try 'cd build && cmake ../' ")
 endif(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
 cmake_minimum_required(VERSION 2.8)
 project(gnss-sdr CXX C)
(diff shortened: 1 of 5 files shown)


carlesfernandez:
Merged at 'next'. Thanks Anthony!

PR #8, opened 2015-04-30
OsmoSdr Flags

anthony-arnold:
This is a change to introduce a new SignalSource config string known as `SignalSource.osmosdr_args`. These arguments will be sent to the GrOsmoSdr library when building the `osmosdr::source` object of the `osmosdr_signal_source` adapter. 

Sending arguments to this constructor allows greater control over the library, including specifying which device to use (in case you have more than one plugged in) and other fun stuff such as connecting to an `rtl_tcp` host over a network.

See http://sdr.osmocom.org/trac/wiki/GrOsmoSDR for details.

The default value for this config string is just an empty string, which is the default argument for the `osmosdr::source` constructor. So old behaviour won't be affected.

I performed a test by running `rtl_tcp -a <network address> -f 1575420000 -g 0 -s 2000000` on one computer  attached to a LAN and with an RTL dongle attached. I then ran GNSS-SDR on a separate computer attached to the same LAN with the configuration value set to `SignalSource.osmosdr_args=rtl_tcp=<address of remote computer>`. GrOsmoSdr connected just fine, and started reading I/Q data of the LAN.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -444,7 +444,7 @@ if (NOT GFlags_FOUND)
      ExternalProject_Add(
           gflags-${gflags_RELEASE}
           PREFIX ${CMAKE_CURRENT_BINARY_DIR}/gflags-${gflags_RELEASE}
-          GIT_REPOSITORY https://github.com/gflags/gflags
+          GIT_REPOSITORY git://github.com/gflags/gflags.git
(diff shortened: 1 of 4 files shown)


carlesfernandez:
As always, great contribution! Thanks Arnold :-)

PR #14, opened 2015-10-14
Two bit packed file source

odrisci:
This pull request adds a generic two bit packed file source. The data is assumed to be packed as bytes ```item_type=byte``` or shorts ```item_type=short``` so that there are 4 two bit samples in each byte.

The two bit values are assumed to have the following interpretation:

| b_1 | b_0 | Value |
|------ |------|----------|
|   0   |   0  |   +1     |
|   0   |   1  |   +3     |
|   1   |   0  |   -3      |
|   1   |   1  |   -1      |

Within a byte the samples may be packed in big endian ```big_endian_bytes=true``` or little endian ```big_endian_bytes=false``` order. If the order is big endian then the most significant two bits will form the first sample output, otherwise the least significant two bits will be used.

Additionally the samples may be either real ```sample_type=real```, or complex. If the sample type is complex, then the samples are either stored in the order: real, imag, real, imag, ... ```sample_type=iq``` or in the order: imag, real, imag, real, ... ```sample_type=qi```.

Finally, if the data is stored as shorts ```item_type=short```, then it may be stored in either big endian ```big_endian_items=true``` or little endian ```big_endian_items=false```. If the shorts are big endian then the 2nd byte in each short is output first.

The output data type is either ```float``` or ```gr_complex``` depending on whether or not ```sample_type``` is ```real```.

A sample config file is also provided, which simply reproduces the GPS L1 NSR config file using the new (more general) file source. I haven't been able to test this particular configuration as I don't have an NSR, so it would be good if someone could try this out.

This has been tested on a sample source with the following configuration:
```
;######### SIGNAL_SOURCE CONFIG ############
;#implementation: Use [File_Signal_Source] or [UHD_Signal_Source] or [GN3S_Signal_Source] (experimental)
SignalSource.implementation=Two_Bit_Packed_File_Signal_Source

;#filename: path to file with the captured GNSS signal samples to be processed
SignalSource.filename=/data/150924-1450.datz

;#item_type: Type and resolution for each of the signal samples. Use only gr_complex in this version.
SignalSource.item_type=short

;#sampling_frequency: Original Signal sampling frequency in [Hz]
SignalSource.sampling_frequency=60000000 

;#freq: RF front-end center frequency in [Hz]
SignalSource.freq=1575468750 

;#samples: Number of samples to be processed. Notice that 0 indicates the entire file.
SignalSource.samples=6000000000

;#repeat: Repeat the processing file. Disable this option in this version
SignalSource.repeat=false

;#dump: Dump the Signal source data to a file. Disable this option in this version
SignalSource.dump=false

SignalSource.dump_filename=./signal_source.dat


;#enable_throttle_control: Enabling this option tells the signal source to keep the delay between samples in post processing.
; it helps to not overload the CPU, but the processing time will be longer.
SignalSource.enable_throttle_control=false

SignalSource.sample_type=iq
SignalSource.big_endian_items=true
SignalSource.big_endian_bytes=false
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *~
+.*.swp
 docs/doxygen/Doxyfile
 docs/html
 docs/latex
--- a/conf/gnss-sdr_GPS_L1_nsr_twobit_packed.conf
+++ b/conf/gnss-sdr_GPS_L1_nsr_twobit_packed.conf
@@ -1,0 +1,152 @@
+; Sample configuration file for IFEN SX-NSR software receiver front-end
+; http://www.ifen.com/products/sx-scientific-gnss-solutions/nsr-software-receiver.html
+; This sample configuration is able to process directly .sream binary files
+; You can define your own receiver and invoke it by doing
+; gnss-sdr --config_file=my_GNSS_SDR_configuration.conf
+;
+
+[GNSS-SDR]
+
+;######### GLOBAL OPTIONS ##################
+;internal_fs_hz: Internal signal sampling frequency after the signal conditioning stage [Hz].
+GNSS-SDR.internal_fs_hz=2560000
+
+;######### CONTROL_THREAD CONFIG ############
+ControlThread.wait_for_flowgraph=false
+
+;######### SUPL RRLP GPS assistance configuration #####
+GNSS-SDR.SUPL_gps_enabled=false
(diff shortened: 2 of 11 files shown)


carlesfernandez:
Merged into next. Thank you Cillian! :-)

PR #16, opened 2015-11-15
update version of armadillo in CMakeLists.txt

sundw2014:
The old armadillo is not available


carlesfernandez:
Merged at 'master'.

PR #20, opened 2016-02-08
Beidou GSoC 2015

giorgiosavastano:
BeiDou signal generation and acquisition by Giorgio Savastano


carlesfernandez:
Hi Giorgio,

the pull request has some issues that must be resolved before merging to next. Tests require the generation of files with Matlab scripts (for a good reason, they are 12 MB each), but they cannot be added to the repository due to its large size. Why don't you generate the test signals with you own developed Beidou signal generator?

In addition, please pull from next and **do not modify** non-related Beidou files, apart from CMakeLists.txt files.

Best regards,
Carles

PR #30, opened 2016-05-28
Fix usage of volk_32f_index_max*, first argument ...

michaelld:
... , to be uint16_t* as required by the Volk API. Simple fix.


carlesfernandez:
Thanks Michael for the fix. This is an issue since it breaks building if you are using a Volk install prior to https://github.com/gnuradio/volk/commit/cb83e1ce4280f84d85eb7d7b6c17179b57f20d98 (for instance, through the libvolk-dev 1.2.2-1 package in Debian).

Opened an issue at https://github.com/gnss-sdr/gnss-sdr/issues/31

michaelld:
Yes, that is certainly an issue. No good way to check for it either except via some clever CMake code. If Volk were at 1.2.3, then that would be a simple check. Otherwise, testing the API is probably the best way to go. What a PITA.

Issue #31, opened 2016-05-29
Change in VOLK's API breaks building


carlesfernandez:
The interface of  volk_32f_index_max_16u have changed as per https://github.com/gnuradio/volk/commit/cb83e1ce4280f84d85eb7d7b6c17179b57f20d98 . This breaks building if GNSS-SDR is using Volk built from the latest source code snapshot. However, Volk’s master branch version is still tagged as ‘1.2.2’, so there is no way to detect this API change (for instance if Volk was installed through the libvolk-dev package, or from the ‘1.2.2’ tag.)

carlesfernandez:
Opened an issue at https://github.com/gnuradio/volk/issues/84

carlesfernandez:
Workaround at 0aa980a9a10a6d7bdb01c8cdac7a4cf2b4e9a028 

However, this will break building of code tagged with v0.0.7 as soon as VOLK 1.2.3 is released

Alex--C:
The workaround should be included in the "next" branch, right?

I'm getting the following error when I try to build it:

..../src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition_cc.cc:339:84: error: cannot convert 'unsigned int*' to 'uint16_t* {aka short unsigned int*}' in argument passing
		      volk_32f_index_max_16u(&indext, d_magnitude, effective_fft_size);

latinovic:
On Debian gnss-sdr version 0.0.7-1 mentioned workaround is not included so this behavior is expected, but I had tried latest version from git repository and build failed with the same error as well.
I investigated further and it seems that  VOLK_VERSION stuck on 1.1.
The reason for that is in cmake/Modules/FindVolk.cmake:
**set(VOLK_VERSION ${PC_VOLK_VERSION})**

After commenting this line I was able to build gnss-sdr successfully.

odrisci:
This fix caused some problems for me using the Galileo E1 pcps acquisition with a very high sampling rate. It turns out that the 16 bit index isn't enough. For example, with 50 Msps, there are 200 000 samples in every E1b code period,  but the current fix will only look in the first 65 535 (2<sup>16</sup>-1) samples for the maximum value. The situation would be much worse for E5, where the code period is 10 ms.

I've created a pull request (#34) that addresses this by using the `volk_32f_index_max_32*` functions and 32 bit unsigned indexing.

carlesfernandez:
[...]

PR #34, opened 2016-08-30
Alternative fix for issue #31

odrisci:
Rather than changing the index type to uint16_t, I think it would be
better to use the volk_32f_index_max32* functions instead of the 16 bit
counterparts. This ensures backwards compatibility (the index was
previously 32 bit, even if the function name indicated that it was 16
bit) and also, for FFT acquisition we may encounter large FFTs with more
than 65 535 points


odrisci:
Wait, will need to update this. The `volk_32f_index_max32*` functions were only introduced in Volk gnuradio/volk@2878d1aa. Need to modify my fix

Issue #69, opened 2017-09-11
Proposal for non-blocking acquisition


odrisci:
As the sampling rate and/or frequency uncertainty increases the acquisition block becomes the bottleneck in performance. This is a problem for real time operation with Galileo (at least 5 Msps) or with front-ends with poor oscillators (like the HackRF, which has a 10 ppm oscillator, leading to a frequency uncertainty in the +/- 15 kHz range).

The idea is to modify the `pcps_acquisition_cc` block so that the bulk of the work is done in a separate thread. A new parameter `blocking` is passed at construction, if this is `true` the behaviour is as normal, if its `false` then the acquisition block skips over samples that arrive while the processing thread is busy. The default value is `true`, as the existing implementation is blocking.

This I think is what the `pcps_multithread_acquisition` block was designed for, but this never worked for me. This implementation has a couple of advantages  in that it re-uses the existing `pcps_acquisition_cc` block and it uses standard C++11 threading concepts.

The disadvantage of this approach is that we introduce yet another thread, with the associated overhead that comes with that. However, I've found that this enables real-time operation that would otherwise not be possible as all threads are held up waiting for the acquisition block to process.

I've pushed a working implementation to my fork [here](/odrisci/gnss-sdr/tree/non_blocking_acq). Comments are welcome.

carlesfernandez:
Dear @odrisci ,

this was exactly the idea behind `pcps_multithread_acquisition`, but we were just unable to make it play well with the GNU Radio schedule. You did, and it is even integrated into `pcps_acquisition_cc`. Very cool!

I'm quite clogged these days with other duties, but I'll try to test it during the weekend with complex configurations. May I have your permission to merge / add it in a new branch in the upstream repository ?

Thank you very much for looking into this.

odrisci:
Hi Carles

I've created a pull request for this at #72 - if you have a chance to test it yourself that would be great. I have been using it in real time operation over a number of hours using both a BladeRF and a HackRF and it seems to be fine, but it would be good to have some more tests.

Cillian

carlesfernandez:
Ok, I did a new test:

taking as a input the file 2013_04_04_GNSS_SIGNAL_at_CTTC_SPAIN.dat, available at https://sourceforge.net/projects/gnss-sdr/files/data/ 

I tested the following configuration:

```
[GNSS-SDR]

;######### GLOBAL OPTIONS ##################
GNSS-SDR.internal_fs_sps=4000000

[...]

PR #72, opened 2017-09-14
Added 'blocking' option to pcps_acquisition_cc

odrisci:
Using C++ 11 threading routines to create a worker thread that can
perform the FFT for acquisition in a non-blocking manner. By default
'blocking' is set to true, which is identical to the previous behaviour
(with the added overhead of an extra thread). When 'blocking' is set to
false, then the main thread does not wait for the worker, thereby
letting the receiver continue even when the FFT thread is working. This
is very useful for using FFT acquisition at higher sampling rates, where
the FFT cannot be performed in real-time.

The aim is to replace the separate pcps_multithread_acquisition_cc.

TODO: implement the same thing for the pcps_acquisition_sc

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,11 +584,10 @@ We use a [DBSRX2](https://www.ettus.com/product/details/DBSRX2) to do the task, 
     1. The default configuration file resides at [/usr/local/share/gnss-sdr/conf/default.conf](./conf/gnss-sdr.conf).
     2. You need to review/modify at least the following settings:
         * ```SignalSource.filename=``` (absolute or relative route to your GNSS signal captured file)
-        * ```GNSS-SDR.internal_fs_hz=``` (captured file sampling rate in Hz)
-        * ```SignalSource.sampling_frequency=``` (captured file sampling rate in Hz)
-        * ```SignalConditioner.sample_freq_in=``` (captured file sampling rate in Hz)
-        * ```SignalConditioner.sample_freq_out=``` (captured file sampling rate in Hz)
-        * ```TelemetryDecoder.fs_in=``` (captured file sampling rate in Hz)
+        * ```GNSS-SDR.internal_fs_sps=``` (captured file sampling rate in samples per second)
+        * ```SignalSource.sampling_frequency=``` (captured file sampling rate in samples per second)
+        * ```SignalConditioner.sample_freq_in=``` (captured file sampling rate in samples per second)
+        * ```SignalConditioner.sample_freq_out=``` (captured file sampling rate in samples per second)
     3. The configuration file has in-line documentation, you can try to tune the number of channels and several receiver parameters. Store your .conf file in some working directory of your choice.
 4. Run the receiver invoking the configuration by
 ```$ gnss-sdr --config_file=/path/to/my_receiver.conf```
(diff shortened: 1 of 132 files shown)


carlesfernandez:
It works well in runtime, but tests `GpsL1CaPcpsAcquisitionTest.ValidationOfResults` and `GpsL2MPcpsAcquisitionTest.ValidationOfResults` fail. Other tests using the same block (such as `GpsL1CaPcpsAcquisitionGSoC2013Test.ValidationOfResults`) succeed, so maybe is a problem of the test. Looking into it.

carlesfernandez:
Ok, I think I’ve got it.

It seems that the `blocking=false` option is not exactly non-blocking. This can be seen in  https://github.com/carlesfernandez/gnss-sdr/tree/odrisci-non_blocking_acq-test, where I did the following:

* Modified gps_l1_ca_pcps_acquisition_test.cc in order to wait for the Acquisition block to emit a message, and then stop the top_block. 
* Left galileo_e1_pcps_ambiguous_acquisition_test.cc as is (i.e., without any “waiting mechanism”)

The result is that, in my machine, `GpsL1CaPcpsAcquisitionTest` always passes but `GalileoE1PcpsAmbiguousAcquisitionTest` **sometimes passes, but sometimes it does not**.  In my interpretation, it just depends on how fast the acquisition_core thread returns the result message. The GNU Radio schedule just ask for more data, but the acq block is busy and let samples flow downstream. If the end of the data file is found before the acquisition block emits a message, the result is a `0` message (“Stop the receiver”) instant of `1` (“Positive Acquisition”).

This breaks the determinism provided by the GNU Radio schedule, since the result of the test depends on how fast the host machine performs the required signal processing. Even in the same machine, the test fails from time to time (!!).

I’m reluctant to merge the code as is into `next` because of this issue. Maybe it could be a new GNU Radio block `pcps_acquisition_non-blocking_cc.cc` called by the adapters according to the `blocking`  parameter. I’m willing to work on this, but I’m not sure if this would be the best way to go. 

Any other ideas on how to meet the GNU Radio schedule requirements with this approach?

odrisci:
Ok. I see it now.  My logic is flawed.  I'll need to rework it a bit so that the calls to `send_positive_acquisition` etc occur in the worker thread. I'll work on it over the next few days and push an update. 

Thanks for your work on the tests.  I'll incorporate this into my update

odrisci:
Ok, I've added some updates now - it passes the tests on my machine all the time now...

I haven't added a test for the case where `blocking=false`, we should probably add this before any potential merge. Also, I presume we would need to rebase from `next`.

Can you check that my updates work on your machine as well?

carlesfernandez:
Ok, now it's working for me. I've moved to a branch called [`non_blocking_acq`](https://github.com/gnss-sdr/gnss-sdr/tree/non_blocking_acq) in the upstream repo. It's in sync with `next`, and I will push there any further commit to `next`, so it will be ready to merge when done. 

The increase in speed is awesome. Thanks for your contribution!

carlesfernandez:
I'm closing this pull request, so we can continue the conversation in https://github.com/gnss-sdr/gnss-sdr/issues/69

PR #80, opened 2017-10-11
GLONASS Navigation Update

dmiralles2009:
This adds a couple of settings needed for RTKLib in order to properly parse GLONASS measurements. In addition, it fixes a couple of bugs with the time computation and proper pseudoranges measurements are now reported by the software. Finally, this is done in order to keep the rest of the team with the most updated version of the code


carlesfernandez:
👍

PR #111, opened 2018-01-23
Modify dump file pcps acquisition

antonioramosdet:
Migrate acquisition dump files to .mat


carlesfernandez:
This commit breaks functionality (e.g., the test:

$ $ ./run_tests --gtest_filter=GpsL1CaPcpsAcquisitionTest* --plot_acq_grid

does not work anymore. We should not remove prior functionality. To be discussed.

antonioramosdet:
Improving compatibility with prior versions

antonioramosdet:
Removed Matio min. required version. Now there are not conflicts with older versions of Matio library

PR #137, opened 2018-03-13
bugfix: Fix acquisition test for GLONASS

Gastd:
The acquisition test was failing in my computer. The expected Doppler was wrong.


carlesfernandez:
I am curious about this test. In MacOS, the test passes with -8000 and fails with -9500. In Linux (many distributions tested!), it is the other way around. I accepted this PR since Linux is our main targeted OS, but I'm trying to figure out why this is happening.

PR #152, opened 2018-04-11
Change std::deque to boost::circular_shift

antonioramosdet:
This modification increases notoriously the performance of the observables block. For instance, see the execution time difference in the HybridObservablesTest.ValidationOfResults test


antonioramosdet:
Solving the conflicts

PR #167, opened 2018-05-08
Added a config file for BDS B2a

sarahrbek:


carlesfernandez:
Thanks! I'm still not merging this file, since the blocks are not there, but the pull-request is correctly done.

PR #183, opened 2018-06-13
Kf

glamountain:
Added acquisition synchro parameter for passing doppler bin width from acquisition to tracking. In tracking, this value is used to set the initial state estimation covariance parameters.


glamountain:
Cancelled this pull request. Changes to be submitted w/o merge to next branch

PR #201, opened 2018-08-20
Fixed OS X building

rventura:
Minor changes to allow building on OS X.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ if(ENABLE_PACKAGING)
 endif(ENABLE_PACKAGING)
 
 # Testing
-option(ENABLE_UNIT_TESTING "Build unit tests" ON)
+option(ENABLE_UNIT_TESTING "Build unit tests" OFF)
 option(ENABLE_UNIT_TESTING_EXTRA "Download external files and build extra unit tests" OFF)
 option(ENABLE_SYSTEM_TESTING "Build system tests" OFF)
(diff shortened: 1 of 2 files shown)


carlesfernandez:
Dear @rventura 

Building in MacOS X works well when using -DENABLE_UNIT_TESTING=ON. 

If you are getting errors in the building process, please open a new issue and provide details about the errors and the specific version you are working with.

Thanks,
Carles.

rventura:
Building with unit testing fails because of lack of development files in OS X (at least when gtest is installed via homebrew). I'll open an issue on that.

Not however the fix in rtklib.h that was preventing compilation.

carlesfernandez:
Mmmm now I'm realizing that you are using homebrew and I'm using Macports, that could be the reason why I'm not able to reproduce this issue. 

The problem with -DENABLE_UNIT_TESTING=ON could be related to the fact that GNSS-SDR uses the googletest sources (as recommended) instead of pre-built libraries. If you are not using googletest anywhere else, you can safely remove that package and GNSS-SDR's build system will download googletest source code for you if it is not detected in your system. Or simply doing `cmake -DENABLE_UNIT_TESTING=OFF ..`

rventura:
You are absolutely right. I spotted today the problem. There was a gtest installation from a non-standard brew tap (ROS actually). I unlinked it and now gnss-sdr compile just fine (it downloads and builds googletests as expected).

PR #203, opened 2018-08-23
Fix build issue in OS X

rventura:
Removes #define of thread_t that was colliding to same define in system include in OS X.


carlesfernandez:
Dear Rodrigo,

I'm going to accept this PR, since it is harmless for other OS and it seems to solve your issue. But I'm curious because my daily work is on MacOS and I've never seen this error. Could you please report the version of your OSX?

Thanks,
Carles

rventura:
Sure. It's El Capitan 10.11.6 (15G22010).

PR #210, opened 2018-09-10
Split kf tracking block into separate blocks for each kf implementation

glamountain:
The KF tracking block uses internal logic based on .conf values to determine which "mode" to operate in, however a cleaner implementation can be achieved by splitting this into separate blocks which can be selected by the `Tracking_1C.implementation` parameter. This will make adding new and increasingly distinct KF based tracking blocks a simpler and cleaner process.


glamountain:
This will need to be revised to structure the code as a single kalman filter gnuradio tracking block modified by multiple adapters, in the same way that the `dll_pll_veml_tracking` block consolidates the various dll_pll architecture blocks, in according with the gnss-sdr design. In addition, the methodology for discriminator free tracking of gps l1 signals will need to be reconsidered. Pull request canceled pending future updates.

PR #211, opened 2018-09-11
Add carrier phase rate variation in the carrier wipeoff block

antonioramosdet:


carlesfernandez:
This seems to break the computation of position: https://gitlab.com/gnss-sdr/gnss-sdr/pipelines/29973533

antonioramosdet:
Now the remnant carrier phase is not computed in the Carrier_wipeoff_multicorrelator_resampler function. Instead it is computed in the lines 840 and 841 of dll_pll_veml_tracking.cc. It should be interesting to try the tests again in Gitlab.

carlesfernandez:
Much better: https://gitlab.com/gnss-sdr/gnss-sdr/pipelines/30135701

PR #216, opened 2018-10-08
Fix code for Boost 1.67

giomasce:
Hi!

GNSS-SDR does not compile with recent versions of Boost (I am using 1.67 at the moment). The two attached patches should fix that, hopefully without changing the program behaviour. Could you please confirm me they are correct?

Thanks, Giovanni.


carlesfernandez:
As stated in https://github.com/gnss-sdr/gnss-sdr/blob/master/CONTRIBUTING.md and in https://gnss-sdr.org/contribute/ , we only take pull requests to the ‘next’ branch. The changes you propose were already applied there.

The merging of “next” into “master” will be done in the following weeks.

Thanks for your PR.

PR #261, opened 2019-03-26
BeiDou B3I bug fixes

dmiralles2009:
Some bug fixes to increase accuracy in the BeiDou B3I and B1I position solution accuracy. In addition, some new pushed code broke the BeiDou B3I position solution :(, it will be great if we added a PVT test for it to avoid this in the future. At the moment, I am working to get several unit test added to the code but the best fix will be to run a PVT solution if possible.

Data can be found [here](https://drive.google.com/open?id=1XXo6O0o9cOiFAQytxowGq7rwORnw1R5S) . Simply download the `SJTU_Bands-BD3B3I.dat` file and run the [ION Metadata Standard executable](https://github.com/IonMetadataWorkingGroup/GNSS-Metadata-Standard) as
```
./Converter 'SJTU_Bands-BD3B3I_working.sdrx' 'int8_t'
```
When the file gets converted to the desired type, use the following configuration to run the file
```
$ gnss-sdr --config_file=conf/gnss-sdr_BDS_B3I_byte.conf
```


carlesfernandez:
Hi @dmiralles2009, 

it seems that clang-format didn't read the .clang-format properly and changed the formatting settings, so the result is a little messy. Could you please revert ac984bd ? I would like to spot where is that recent change that broke the code, but the format changes make it hard. Do you know in which file it was?

Thanks,
Carles

carlesfernandez:
@dmiralles2009,

I can cherry-pick the other commits, but please tell me where the newly introduced bug was :-)

dmiralles2009:
Hey @carlesfernandez , it is done. I reverted 1 commit to avoid the clang-format issues I was having. Not sure what happened with my PC there.

The bug in question happened in file beidou_b3i_telemetry_decoder_gs.cc around line 65 to 70. It seems like some messages were left out. I noticed an issue when running the code after getting the latest changes from next.

Anyway, the issue is now solved and I will be working on the unit test over the next few days. I have a bug in the acquisition unit test that is driving me nuts :), but I am planning in also adding unit tests for tracking, decoding, etc. If we use the data file we could check the PVT for the B3I receiver.

Hope this helps. 
Best,
Damian

carlesfernandez:
Ohh, I see. We forgot to enable that message port. Thank you so much for spotting this. I'm working on including Beidou test in the CI system, it's great to have that file for testing. Thanks again!

More on this: commit e7b9f93 contains a couple of binary fines, approx size of 1.5 MB. I really would prefer to move them somewhere else, make the test "EXTRA", and downloading them at configuration time if that option is enabled. I will take care of that, but I cannot accept that commit. If you don’t mind, I will manually get the changes from the rest of the files of that commit, and cherry-pick the previous ones.

dmiralles2009:
@carlesfernandez  No problem at all. Was not aware of the issue with the file size and I thought they were small as they are. Nevertheless, I understand, apologies for making this troublesome to you.

Best,
Damian

carlesfernandez:
Merged at `next` in 1cf49f461ae85451644d085a6da82f215f581e66

Thanks for your contribution!

PR #306, opened 2019-08-26
GSoC 2019 - Optimal frame synchronization detection

lucasventurar:
### Unit test
**gps_l1_ca_telemetry_synchronization.cc**
It has two tests that take into account all the state machine, one with Hard Correlation and the other one with Simplified-LRT (SLRT). Both tests save all the necessary metrics into a csv file to compare them later in terms of synchronization error in the differents states and final synchronization.

The first lines are the necessary includes and then there are 3 constants defined:

- _vector_size_ is the number of samples in the test, is set to 6000 (20 sub-frames with 300 bits). It should be noted that a navigation message has 37500 samples (25 frames of 5 subrames each with 300 bits).

- _preamble_offset_ is the number of samples before the first preamble. Is set to 200, so at 500 the second preamble is added.

- _Nw_ is the number of Monte-Carlo realizations. The test were done with Nw = 10^6.

_GpsL1CATelemetrySynchronizationTest_ is the fixture for testing class. Initializes the variables needed for both tests like in _gps_l1_ca_telemetry_decoder_gs_ and declares three new functions and two vectors:
- _preamble_samples()_: Initializes _d_preamble_samples_ with the SW.

- _make_vector()_: Fills _initial_vector_ with the preamble and random data.

- _fill_gnss_synchro()_: Fills _synchro_vector_ (_initial_vector_ + noise with a standard deviation)

The **HardCorrelator** test simulates the _gps_l1_ca_telemetry_decoder_ frame synchronization with a hard correlation in the state machine. Frist, it creates a document (_synchronization_SC_test_x.csv_) where everything will be saved. Secondly, a for loop is created for the Monte-Carlo realizations. For every realization, the metrics and variables are set to 0, the vectors are emptied, some values are checked and noise is added if need it. Then, the replica of general_work starts, reproducing the state diagram seen before. Depending on the state and some conditions, the following variables are evaluated and filled:

- _n_preamble_detections_s0_: Number of detected preambles in state 0.

- _n_preamble_detections_s1_: Number of detected preambles in state 1.

- _n_correct_detections_s0_: Number of correct detected preambles in state 0.

- _n_wrong_detections_s0_: Number of wrong detected preambles in state 0.

- _n_correct_detections_s1_: Number of correct detected preambles in state.

- _n_wrong_detections_s1_: Number of wrong detected preambles in state 1.

- _n_preambles_: Number of total correct preambles (missed and detected).

- _final_synchronization_: Is set to 0 if no final synchronization is achieved, 1 if correct, 2 if wrong.

The **_SoftCorrelator_** test is similar to the _HardCorrelator_ but implements the SLRT correlation instead of the hard correlation. Saves the metrics in a file called _synchronization_SLRT_test_3.csv_. Lastly, before the next Monte-Carlo realization, the results are stored and a condition is checked. Finally, when all the realizations are complete, the file is closed.

The next stage is to compare both correlators with the saved data.

### Improved telemetry decoder
The new block implements different correlations. If nothing is defined or if __correlation_ is defined as “HC”, it performs as before, doing the hard correlation. But if _correlation is defined as “SC” or “SLRT” it does the soft correlation and Simplified-LRT.

Once it enters the frame synchronization, depending on the parameter __correlation_ the correlation is computed differently:

- HC: abs(_corr_value_) >= _d_samples_per_preamble_.

- SC: abs(_corr_value1_) >= λ.

- SLRT: abs(_corr_value1_) – _corr_value2_ >= λ.

Where the differents correlation values are computed like this:

- _corr_value_ += sign(_d_symbol_history[i]_) · _d_preamble_samples[i]_

- _corr_value1_ += _d_symbol_history[i]_ · _d_preamble_samples[i]_

- _corr_value2_ += abs(_d_symbol_history[i]_)


carlesfernandez:
Please resolve the merging conflicts.

carlesfernandez:
Closing this PR by lack of activity

PR #327, opened 2019-11-05
Add clock drift

odrisci:
I work with a lot of low cost front ends which often have a large clock drift (usually measured in parts per million). It would be very useful to be able to extract this information from gnss-sdr. This pull request attempts to do this.

Unfortunately, the RTKLIB engine does not store the clock drift in its output structure, even though the drift is naturally computed as part of the velocity solution (in the same way as clock bias is computed as part of the position solution). So I've used the previously unused field `pvt_sol.dtr[5]` to capture the drift as it is computed in RTKLIB (in unts of m/s). 

I've added `clock_drift_ppm` field to the `RtklibSolver` class and also added `user_clk_drift_ppm` to the `monitor_pvt` class. I've added this as the last element in the hope that the serialization/deserialization won't be an issue across different versions (at least the protobuf version should be Ok, not so sure about Boost serialization). Note for the gnss-sdr variables I'm storing the drift in parts per million as I think this is the natural choice, but we can easily revert to m/s to be in keeping with RTKLIB or s/s to be in keeping with the clock bias fields.

### DIFF
--- a/docs/protobuf/monitor_pvt.proto
+++ b/docs/protobuf/monitor_pvt.proto
@@ -38,4 +38,6 @@ double gdop = 25;  // Geometric Dilution of Precision
 double pdop = 26;  // Position (3D) Dilution of Precision
 double hdop = 27;  // Horizontal Dilution of Precision
 double vdop = 28;  // Vertical Dilution of Precision
+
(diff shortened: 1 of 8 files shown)


carlesfernandez:
Awesome! Thanks Cillian, I'll merge it ASAP.

Issue #331, opened 2019-11-08
Bad normalization in pcps_acquisition


odrisci:
The pcps acquisition block is a generic block providing non-coherent parallel code phase search. At the moment the normalization of the decision variable is incorrect (probably my fault when I added the non-blocking feature). For this reason it is not possible (at least in my experience) to write a generic configuration file for acquisition - each file needs to be tuned for the data set it processes in order to set the acquisition threshold in an appropriate manner.

The solution is fairly straightforward: the normalization can be computed by averaging the total power in a Doppler bin removed from the one containing the peak. The peak decision statistic is then divided by this power for a correct normalization.

There is one more issue then, that the current method of computing the decision threshold is based on the assumption that there is no non-coherent accumulation (max_dwells == 1). In fact it is straightforward again to extend the computation of the threshold to cases where max_dwells > 1.

In putting together a fix for this I've come across another issue: code duplication. In order to implement the fix for the threshold computation I have to modify every one of the individual adapters that adapts the pcps_acquisition class, and there are many. Most of this code is identical but has to be re-written for each class, making it very bug prone and hard to maintain.

I would suggest changing the structure of the acquisition classes a little bit as follows:

  1. The acq_conf class should be given a member function to set all its variables from a ConfigurationInterface. This would cut down on lots of the boilerplate in the acquisition adapaters (the same should be done for the dll_pll_conf class)
  2. The threshold computation should be done inside the gnuradio block (pcps_acquisition in this case) rather than in the adapter. I think this makes more sense as the gnuradio block implements the actual acquisition algorithm and so should have the knowledge on how to best compute the threshold.

carlesfernandez:
Fix merged into the next branch at 82508e9aba50e56d0abfe3374ef3f0627eec5b90

PR #332, opened 2019-11-08
Fix pcps threshold

odrisci:
[DO NOT MERGE YET!]

This is a relatively big pull request implementing the fixes discussed in #331 .

It is open for discussion as there are some changes in here that I'm not sure are exactly the way to go, but they do start to address the issue of code duplication in the pcps_acquisition adapters, and also apply the fix for the pcps acquisition threshold.

### DIFF
--- a/src/algorithms/acquisition/adapters/beidou_b1i_pcps_acquisition.cc
+++ b/src/algorithms/acquisition/adapters/beidou_b1i_pcps_acquisition.cc
@@ -200,15 +200,15 @@ float BeidouB1iPcpsAcquisition::calculate_threshold(float pfa)
 
 void BeidouB1iPcpsAcquisition::connect(gr::top_block_sptr top_block)
 {
-    if (item_type_ == "fc32")
+    if (item_type_ == "gr_complex")
         {
             // nothing to connect
         }
-    else if (item_type_ == "ic16")
+    else if (item_type_ == "cshort")
         {
             // nothing to connect
         }
-    else if (item_type_ == "ic8")
+    else if (item_type_ == "cbyte")
         {
             // Since a byte-based acq implementation is not available,
             // we just convert cshorts to gr_complex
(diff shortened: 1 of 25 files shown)


carlesfernandez:
I understand the key issue here is at  75c57e90e5fbec1e5ede6db7dfd11d4f962a700e I like it. See #336 about how we do proceed about removing code duplication and moving to a new data type nomenclature. All those things mixed together are hard to process!

odrisci:
So I've reverted to the gnss-sdr style naming schemes here, and also added some unit tests for the item type helpers

I do have a question on the difference between "ibyte" and "cbyte". It seems to me that the memory layout is identical in each case. The only difference is that the item size is one byte for ibyte and 2 bytes for cbyte. I don't see the use case for ibyte - the only difference is when you expect your source stream to send you an uneven number of samples (one extra in-phase sample), but I would prefer to always ensure that I got a matching Q sample for every I. Similar for "ishort". Surely it would be better to just enforce use of "byte" or "short" for real types and "cbyte" or "cshort" for complex types. Or am I missing something?

carlesfernandez:
Hi @odrisci 

you are right that "ibyte" and "ishort" make little sense, it is just a way to indicate that they are real samples. Ok to rename them to "byte" and "short" (with alias to the old names).

I've been working in this branch at [odrisci-fix_pcps_threshold](/gnss-sdr/gnss-sdr/tree/odrisci-fix_pcps_threshold) and I think it's ready to be merged. It passes all the tests in our CI without breaking existing configurations.

However, there is a change in the configuration. Now the `use_CFAR_algorithm` parameter is not exposed anymore to the configuration file, it is just an internal flag. If the user sets it, the setting is ignored. The usage is as follows:

 * If a manual `threshold` is set by the user (and `pfa` = 0, or not set), the first_vs_second_peak_statistic metric is used.  This is a fixed threshold.

 * If the `pfa` parameter is specified in the configuration file to a non-zero value, then the max_to_input_power_statistic metric is used, and the threshold is set (and recomputed as needed) automatically.

Issue #333, opened 2019-11-08
Incorrect normalization of the non coherent DLL discriminator


odrisci:
The function dll_nc_e_minus_l_normalized only has the correct normalization in the case that the correlation function slope is 1 (BPSK) and the early late spacing is 1 chip (0.5 Early-prompt, or prompt-late).

This should be fixed by adding the early late spacing and correlator slope as parameters and applying the normalization:

```
( 1 - slope * spacing )/slope
```
The same applies to the veml discriminator, but I don't have a nice way of normalizing this as is. If we changed the veml discriminator to:

```
( E^2 + VE^2 - L^2 - VL^2)/(E^2 + VE^2 + L^2 + VL^2)
```
Then this would be easier to normalize. I don't have a good handle on the logic behind the VEML discriminator so I'm not sure if this is a valid approach.

I will submit a pull request for the non-coherent early minus late discriminator

PR #334, opened 2019-11-08
Fix normalization in dll_nc_e_minus_l_normalized

odrisci:
Fix for #333 - this does not change the default behaviour but adds an interface to the dll_nc_e_minus_l_normalized discriminator so that it can be correctly normalized for any correlation function and correlator spacing.


carlesfernandez:
I'm testing this change in the `odrisci-fix_discriminator_normalization` branch, after appling those new parameters in the tracking loops (see commit 970fb6f8482671eed3e2c5c196a16b2359f849d4)

odrisci:
Thanks Carles. I probably should have done that!

odrisci:
I've just pushed a commit incorporating your additions and adding a unit test.

In the process I've realised that the normalization has some problems, so please hold off on merging!

The problem is this: the normalization assumes that the correlation function can be approximated by a straight line with a slope and spacing with an intercept on the y-axis of 1 when the tracking error is zero. The problem is that the y-intercept is not 1 in general, but only for the first and last linear segments of a BOC correlation function (and always for BPSK). So we need to add another parameter to the normalization: the y-intercept, which should default to 1.

I'm working on this too.

carlesfernandez:
I've merged this branch and [odrisci:update_tracking](/odrisci/gnss-sdr/tree/update_tracking) into [gnss-sdr:odrisci-fix_discriminator_normalization](/gnss-sdr/gnss-sdr/tree/odrisci-fix_discriminator_normalization), including the latest additions to next.

I'm starting testing from there.

carlesfernandez:
See #333

PR #350, opened 2019-12-19
For Cross Compiling fix a bug

xusong0zju:
This Commit was important for cross-compile. This was tesed by RK3308(arm64)-Debian10.


carlesfernandez:
How are you cross-compiling? This is usually done by using a toolchain file (see [https://gnss-sdr.org/docs/tutorials/cross-compiling/](https://gnss-sdr.org/docs/tutorials/cross-compiling/)). If this is the case, the variable `VOLK_GNSSSDR_COMPILER`  should be empty.

If you are cross-compiling WITHOUT a toolchain file, then you are right and that variable should take the values passed to cmake. This could be done by:

```
if(CMAKE_CROSSCOMPILING AND CMAKE_TOOLFILE_CHAIN)
        set(VOLK_GNSSSDR_COMPILER "")
    else()
        set(VOLK_GNSSSDR_COMPILER -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER} -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER})
    endif()
```
Does this solution work for you?

carlesfernandez:
The issue should be fixed in 7b4523a, so I'm closing this PR. Please feel free to reopen it if the problem persists.

PR #359, opened 2020-02-14
Added handling of unequal number of input and output channels.

BartSchipperijn:
If a SignalSource block in GNSS-SDR has one input but multiple radio frequency chains (as in the example on https://gnss-sdr.org/docs/sp-blocks/signal-source/#multiple-radio-frequency-chains) the creation of the valve (in gnss_sdr_valve.cc) fails with a coredump. This code change fixed this problem (see also issue 354).


jwmelto:
@BartSchipperijn you might want to review the [contributing guidelines](https://github.com/gnss-sdr/gnss-sdr/blob/main/CONTRIBUTING.md). In particular, your PR should be against **next**, and must be in a branch.

PR #416, opened 2020-07-31
Observables and PVT Files

piyush0411:
Some minor modifications have been made in observables and PVT calculation block.
Most of the work for PVT of Galileo E5b and its overlapping scenarios with other signals was already implemented, only some minor additions were required.
@carlesfernandez @luis-esteve  Please have a look sir.

Thanks and Regards
Piyush


piyush0411:
@carlesfernandez 
There are some configuration files for GNSS receivers in `conf` folder. Should one exist for E5b signal too?
If yes, then which file should be used as signal source in the configuration?

Regards
Piyush

PR #455, opened 2020-12-07
add BDGSA and BDGSV nmea messages

liu000888:
add BDGSA and BDGSV messages for beidou.


carlesfernandez:
Thanks! Merging into next.

PR #468, opened 2021-01-27
Fixed segfault when rinex output is disabled

### LINKED ISSUE #467: Segmentation Fault with RINEX output disabled
After the recent changes in the RINEX printer, a bug has been introduced where a segmentation fault occurs when the PVT block starts iterating for the solution. This is caused by the getter `Rinex_Printer::is_rinex_header_written()` being called even in the case of the instance `d_rp` in rtklib_pvt_gs.cc being unset in the case when RINEX output is disabled.
The segfault does not occur when RINEX output is enabled.

I will submit a pull request with a simple logic fix to avoid this bug.

stefanlinden:
This fixes #467.
A simple logic fix checking whether RINEX output is enabled or disabled before performing the check whether the RINEX header has been written. This avoids a segfault when the RINEX output is disabled.


carlesfernandez:
Thanks! I've extended the fix to other signals than GPS L1.

stefanlinden:
Great! I missed the others since I was testing with GPS L1 data...

PR #475, opened 2021-02-15
Feature/signal source interface

jwmelto:

[comment]: # (
)

[comment]: # (
SPDX-FileCopyrightText: 2011-2021 Carles Fernandez-Prades <carles.fernandez@cttc.es>
)

This PR creates a abstract `SignalSourceInterface`. It also removes a lot of duplicated code into a common base class, `SignalSourceBase`, and further abstracts file-based interfaces behind `FileSourceBase`.

I did my best to compile all the sources, but I'm not convinced all sources even compile anymore. In particular, I was unable to build with `-DENABLE_ARRAY=ON`, `-DENABLE_FLEXIBAND=ON`, or `-DENABLE_GN3S=ON`.


carlesfernandez:
Wow! Thank you @jwmelto for this carefully prepared PR. I'm just starting to review it, but it looks full of very nice ideas!

carlesfernandez:
Hi @jwmelto ,

There is one issue: `std::string_literals` didn’t exist in C++11, so building breaks in old compilers such as GCC 4.8 (CentOS 7, Ubuntu 14.04). Since we still aim to support C++11 and those compilers, I’ve added a workaround that seems to work well and it is not so invasive.

The workaround consists of adding an include file where the `“”s` operator is defined in the `std::string_literals` namespace, only when using C++11. Of course, those compilers raise a warning since literal operator suffixes not preceded by `_` are reserved for future standardization (in this case, it certainly happened), but this is only compiled when using C++11 so it seems legit to me. I’m sure there are more elegant (or safer) ways to do that, but at least this solution seems to work well and the receiver does the right thing in those environments.

For some reason, I can't push commits to this PR. The commit is at https://github.com/carlesfernandez/gnss-sdr/commit/c0e212e44670f1222f6a6bccc8726e0e28d23201 Could you please cherry-pick it and push it to this branch?

jwmelto:
> There is one issue: `std::string_literals` didn’t exist in C++11, so building breaks in old compilers such as GCC 4.8 (CentOS 7, Ubuntu 14.04). Since we still aim to support C++11 and those compilers, I’ve added a workaround that seems to work well and it is not so invasive.

Good point. I knew that user-defined literals were introduced in C++11, but hadn't bothered to worry about the fact that `std::string_literals` were introduced in C++14.

After spending all night fighting with the CI checks, I'm surprised your older environment support isn't part of this? At what point do you stop supporting new development on old environments? Backwards compatibility comes at a cost.

> For some reason, I can't push commits to this PR. The commit is at [carlesfernandez/gnss-sdr@c0e212e](https://github.com/carlesfernandez/gnss-sdr/commit/c0e212e44670f1222f6a6bccc8726e0e28d23201) Could you please cherry-pick it and push it to this branch?

Rather than figure out how to cherry-pick from your repo, I added you to mine. You can push as you see fit.

One note: you removed the `filesystem` support from **signal_source/adapters/CMakeLists.txt** but I didn't see it go anywhere else? I did some surgery along the way (modifying and building my own version of log4cpp to add C++17 support) to force the use of `std::filesystem` (as opposed to `boost::filesystem`), and that change was absolutely required. I don't see the benefit of removing it.

carlesfernandez:
Hi @jwmelto ,

thanks for sending the invitation to your repo. I've just pushed the `std::string_literals` commit there.

Regarding CI, we have some formatting checks and basic building in GitHub Actions, but more exhaustive testing is done in GitLab. at https://gitlab.com/gnss-sdr/gnss-sdr If you check the CI pipelines there, you will see jobs that build gnss-sdr in a list of Linux distributions and versions, then we do some actual tests (with data files) for a set of receiver configurations, and finally, we build and push some Docker images.

Regarding the ' filesystem'  support, I don't see any call to the library within the Signal Source adapters, so that's why I removed it. Anyway, we now have [src/algorithms/libs/gnss_sdr_filesystem.h](https://github.com/gnss-sdr/gnss-sdr/blob/next/src/algorithms/libs/gnss_sdr_filesystem.h). Including that file and linking against `algorithms_libs` will add the support to the library, so you don't need to include the CMake snippet everywhere you need to use ' filesystem` anymore.

jwmelto:
The `<filesystem>` reference is in file_source_base.cc. It uses a filesystem method to get the size of the file, rather than opening and seeking to the end.

The issue I had was, without that snippet (and including gnss_sdr_filesystem.h), the adapters library built against Boost, but the rest of the system built against `std::filesystem`. While it compiled fine, at link time the Boost symbols were undefined.

gnss_sdr_filesystem.h still depends upon the defines provided by the CMake snippet. That's why my comment was that it should be at the top level, so any/all sources that include gnss_sdr_filesystem.h, now or in the future, will compile with the same macros defined.

carlesfernandez:
Ups, I missed that.

Well, if you link against `algorithms_libs`, those macro definitions will still be used (or at least, they are intended to be, since they are in the PUBLIC section of  `algorithms_libs`), so it should consistently use the same library /`std::flielsystem` or `boost::filesystem`, whatever picked up at the root CMakeLists.txt. This was the intention behind `gnss_sdr_filesystem.h`. 

However, it seems that it didn't work for you. Please note that I included `algorithms_libs` in the list of private link libraries for `signal_source_adapters`, so maybe your issue is solved now.

carlesfernandez:
Hi @jwmelto ,

In the GitLab tests, configurations using `Multichannel_File_Signal_Source` were not succeeding. It took me a while to find out the bug: 6d41fc8fb2861efd10678040bf8149a3bd075a8a

I’ve just launched GitLab’s CI again with this new commit. In my local machine, tests are passing successfully again.

jwmelto:
Ugh. I don't like those `std::move` on the argument strings; It's a clang-tidy complaint that is overly paranoid. I also don't like relying on order of initialization, although the language does now define it, so I guess I need to retire my biases.

Your change is correct, and should work. I wonder, though, if in this case it would be better to move the assignment of `rfChannels_` from the initializer list into the ctor body? It's too subtle a dependency, as written. Note, by the way, that if the order of parameters had been swapped in the header, and `rfChannels_` was declared first, this would not have been a failure. Those kinds of nuances are just begging for failure.

Did you want to update the copyright to 2021?

carlesfernandez:
I was betting this was an issue caused by clang-tidy. That particular check is quite evil sometimes.

I respected the copyright year you had in the file, I assumed you started working on it in 2020. Feel free to update it, if you wish. 

All tests are passing now, I think this is ready for merging.

jwmelto:
I am done with it.

carlesfernandez:
I assume `size_t` is `uint64_t` everywhere, right?

jwmelto:
well, at least until 128-bit platforms come out. It really doesn't matter that much which unsigned type you use, as long as it avoids the ignorant ambiguity of the llvm compiler.

So these latest failures ... the MacOS build failure on c221fd1 and the volk-gnsssdr-macos failure on 704b99e ... these are in areas of the code that I should have no impact. I'm not sure what's going on here.

carlesfernandez:
Those are sporadic failures of the CI system. I've been chasing the root of those failures without success. Re-running CI with a new commit with mostly cosmetic changes (see comments).

jwmelto:
Unfortunately, your moving stuff into the base class constructor won't work. You can't call virtual methods from a base class ctor, because the derived class is not fully constructed yet. That's the reason for the init() method in the first place.

jwmelto:
also, it's probably not an official style of this project, but I always try to have method implementations in the same order as the declarations, to help maintainers. That's why the protected ctors were lower in the file than "normal". It's up to you if you want to respect that.

carlesfernandez:
Well, some others like to see the constructor first :hear_no_evil: 

That's a very opinionated issue, I should have not changed that.

carlesfernandez:
ok, merging.

Thanks, @jwmelto, for this awesome work. I especially appreciate the time you spent on those changes that actually change nothing from the user’s perspective but really improve the inner design and facilitate further expansion. You applied here some very nice C++ practices and idioms that I would like to extend to other block families.

acebrianjuan:
Awesome work indeed! :+1: 

@carlesfernandez I think the only thing left now is to add @jwmelto to the to the [list of contributors](https://github.com/gnss-sdr/gnss-sdr/blob/next/AUTHORS).

carlesfernandez:
Thanks @acebrianjuan for the remainder. Done at 8fe9a74ff2a4a9d3ecf2bcc6db764e1df8364ee6

Issue #486, opened 2021-03-10
Fail to check SignalSource.implementation when there are multiple signal sources in conf file


iucgnaw:
When try run gnss-sdr with conf file with multiple signal sources, e.g. gnss-sdr_GLONASS_L1_CA_GPS_L1_CA_ibyte.conf, it fails to check SignalSource.implementation, and exit.

carlesfernandez:
Fixed in #488

PR #488, opened 2021-03-10
Fixed wrong SignalSource.implementation check.

iucgnaw:
Fixed issue: https://github.com/gnss-sdr/gnss-sdr/issues/486

### DIFF
--- a/src/core/receiver/control_thread.cc
+++ b/src/core/receiver/control_thread.cc
@@ -92,23 +92,37 @@ ControlThread::ControlThread()
     const std::string empty_implementation;
     std::string src_impl = configuration_->property("SignalSource.implementation", empty_implementation);
     int src_count = configuration_->property("Receiver.sources_count", 1);
-    if (src_impl.empty() && (src_count != 1))
+    if (src_impl.empty())
         {
-            int num_src = 0;
-            for (int i = 0; i < src_count; i++)
+            if ((src_count == 1))
                 {
-                    std::string src_impl_multiple = configuration_->property("SignalSource" + std::to_string(i) + ".implementation", empty_implementation);
-                    if (!src_impl_multiple.empty())
+                    conf_has_signal_sources_ = false;
+                }
+            else
+                {
+                    int num_src = 0;
+                    for (int i = 0; i < src_count; i++)
                         {
-                            num_src++;
+                            std::string src_impl_multiple = configuration_->property("SignalSource" + std::to_string(i) + ".implementation", empty_implementation);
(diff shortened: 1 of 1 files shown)


src/core/receiver/control_thread.cc
```diff
@@ -92,23 +92,37 @@ ControlThread::ControlThread()
     const std::string empty_implementation;
     std::string src_impl = configuration_->property("SignalSource.implementation", empty_implementation);
     int src_count = configuration_->property("Receiver.sources_count", 1);
-    if (src_impl.empty() && (src_count != 1))
+    if (src_impl.empty())
         {
-            int num_src = 0;
-            for (int i = 0; i < src_count; i++)
+            if ((src_count == 1))
```
jwmelto, line 97: this test is not necessary, and will cause config files with a single "SignalSource0.implementation" section to fail

src/core/receiver/control_thread.cc
```diff
@@ -92,23 +92,37 @@ ControlThread::ControlThread()
...
+                                }
+                        }
+                    if (num_src == src_count)
+                        {
+                            conf_has_signal_sources_ = true;
+                        }
+                    else
+                        {
+                            conf_has_signal_sources_ = false;
+                        }
```
jwmelto, line 119: This whole section can be simplified, by taking advantage of the defined `bool`->`int` conversion:
```c++
            for (auto i = 0; i < src_count; ++i)
                 {
                     auto src_impl_multiple = configuration_->property("SignalSource" + std::to_string(i) + ".implementation", empty_implementation);
                     num_src += !src_impl_multiple.empty();
                 }
             conf_has_signal_sources_ = (num_src == src_count);
```
although I question whether this config file validation is of any value at all, since the flowgraph will fail if the config is mal-formed.

jwmelto (changes requested):
@iucgnaw I ran into this issue earlier this week and was contemplating fixing it.

Take my comments with a grain of salt; I'm not a developer/maintainer, but I'd like to see this (at a minimum) support the following config files:

- a single "SignalSource"
- a single "SignalSource0"
- "SignalSource" followed by "SignalSource1" up to _n_
- "SignalSource0" followed by "SignalSource1" up to _n_

carlesfernandez:
Thanks guys for your contributions. Merging with the changes suggested by @jwmelto 

I've also replaced the parameter name `Receiver.sources_count` by `GNSS-SDR.num_sources` for the sake of consistency. The old name `Receiver.sources_count` is still read, so it does not break compatibility with existing configuration files. If the new parameter exists, it overrides the old one.

PR #495, opened 2021-04-17
Fifo source

lenhart:
Hello,
as suggested by Jim when I asked on the mailing list, I wrote a custom fifo signal source, (the `file_signal_source` is not compatible).

I think the fifo reader is in a state that is acceptable for gnss-sdr, but I am happy to discuss/take feedback on that and adapt things (still quite new to the whole environment). 

Especially interesting are opinions if the performance improves if I read larger chunks from the fifo at once vs. iterating over the items one at a time (which i currently do - I count on the compiler optimization to do some loop unrolling for me, but still..).

Example use cases for the `fifo_signal_source`:
    - multiplex signal streams outside of gnss-sdr
    - another program holds access to the SDR
    - the SDR is not supported by gnss-sdr but can dump the signal to a fifo

Iin separate commits included in this PR I edited a few unrelated files:
- `signal generator`: added missing `std::` to sqrt calls as suggested by clang-tidy
- `file_source_base`: renamed `is_complex_t` to `is_interleaved` to better describe its behavior; modernized c-style cast, fixed typo
- `custom_udp_signal_source`: added ishort support. removed redundant switch block, made overflow events less confusing (`O` is confusable w/ `0`, so I changed it to lower case). Also added overflow explanation to class startup (confused me when I tested the example config file until I checked the code. being more verbose on overflow events is not feasible either as it happens repeatedly).

I also noticed that there are far more signal sources in the code than documented on the website. I was quite fascinated when I discovered some. Is there some decision which ones get mentioned or are the majority just not documented enough? In that case I would (suggest to) add a small list of implementations to the homepage. Even without much description/examples it would be worth it in my opinion.

Cheers!

### DIFF
--- a/src/algorithms/signal_source/adapters/fifo_signal_source.cc
+++ b/src/algorithms/signal_source/adapters/fifo_signal_source.cc
@@ -1,7 +1,7 @@
 /*!
- * \file file_source_base.cc
+ * \file fifo_signal_source.cc
  *
- * \brief Implementation of the class for retrieving samples through a Unix fifo
+ * \brief Implementation of the class for retrieving samples through a Unix FIFO
  * \author Malte Lenhart, 2021. malte.lenhart(at)mailbox.org
  *
  * -----------------------------------------------------------------------------
@@ -18,12 +18,10 @@
 #include "fifo_signal_source.h"
 #include "configuration_interface.h"
 #include "fifo_reader.h"
-#include "gnss_sdr_flags.h"
 #include "gnss_sdr_string_literals.h"
 #include <glog/logging.h>
 #include <gnuradio/blocks/file_sink.h>
 #include <gnuradio/blocks/file_source.h>
-#include <cmath>  // ceil, floor
 
 
 using namespace std::string_literals;
@@ -69,7 +67,7 @@ void FifoSignalSource::disconnect(gr::top_block_sptr top_block)
     if (dump_)
         {
             top_block->disconnect(fifo_reader_, 0, file_sink_, 0);
-            DLOG(INFO) << "disconnected source to file sink";
+            DLOG(INFO) << "disconnected source from file sink";
         }
 }
 
--- a/src/algorithms/signal_source/adapters/fifo_signal_source.cc
+++ b/src/algorithms/signal_source/adapters/fifo_signal_source.cc
@@ -1,0 +1,93 @@
+/*!
+ * \file file_source_base.cc
+ *
+ * \brief Implementation of the class for retrieving samples through a Unix fifo
+ * \author Malte Lenhart, 2021. malte.lenhart(at)mailbox.org
+ *
+ * -----------------------------------------------------------------------------
+ *
+ * GNSS-SDR is a Global Navigation Satellite System software-defined receiver.
+ * This file is part of GNSS-SDR.
+ *
+ * Copyright (C) 2010-2021  (see AUTHORS file for a list of contributors)
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * -----------------------------------------------------------------------------
+ */
+
+#include "fifo_signal_source.h"
+#include "configuration_interface.h"
+#include "fifo_reader.h"
+#include "gnss_sdr_filesystem.h"
+#include "gnss_sdr_flags.h"
+#include "gnss_sdr_string_literals.h"
+#include "gnss_sdr_valve.h"
+#include <glog/logging.h>
+#include <cmath>  // ceil, floor
+#include <fstream>
+#include <utility>  // move
+
+
+using namespace std::string_literals;
+
+FifoSignalSource::FifoSignalSource(ConfigurationInterface const* configuration,
+    std::string const& role, unsigned int in_streams, unsigned int out_streams,
+    Concurrent_Queue<pmt::pmt_t>* queue)
+    : SignalSourceBase(configuration, role, "Fifo_Signal_Source"s),
+      item_size_(sizeof(gr_complex)),  // currenty output item size is always gr_complex
+      fifo_reader_(FifoReader::make(configuration->property(role + ".filename"s, "../data/example_capture.dat"s),
+          configuration->property(role + ".sample_type"s, "ishort"s))),
+      dump_(configuration->property(role + ".dump", false)),
+      dump_filename_(configuration->property(role + ".dump_filename"s, "./data/signal_source.dat"s))
+{
+    if (dump_)
+        {
+            DLOG(INFO) << "Dumping output into file " << (dump_filename_ + ".bin"s);
+            file_sink_ = gr::blocks::file_sink::make(item_size_, (dump_filename_ + ".bin").c_str());
+        }
+
+    if (in_streams > 0)
+        {
+            LOG(ERROR) << "A signal source does not have an input stream";
+        }
+    if (out_streams > 1)
+        {
+            LOG(ERROR) << "This implementation only supports one output stream";
+        }
+}
+
+void FifoSignalSource::connect(gr::top_block_sptr top_block)
+{
+    // here we could add a throttle as done in the file_source_base if required
+    if (dump_)
+        {
+            top_block->connect(fifo_reader_, 0, file_sink_, 0);
+            DLOG(INFO) << "connected source to file sink";
+        }
+}
+
+void FifoSignalSource::disconnect(gr::top_block_sptr top_block)
+{
+    if (dump_)
+        {
+            top_block->disconnect(fifo_reader_, 0, file_sink_, 0);
+            DLOG(INFO) << "disconnected source to file sink";
+        }
+}
+
+size_t FifoSignalSource::item_size()
+{
+    return item_size_;
+}
+
+gr::basic_block_sptr FifoSignalSource::get_left_block()
+{
+    LOG(WARNING) << "Left block of a signal source should not be retrieved";
+    return gr::blocks::file_source::sptr();
+}
+
+
+gr::basic_block_sptr FifoSignalSource::get_right_block()
+{
+    return fifo_reader_;
+}
--- a/src/algorithms/signal_source/adapters/fifo_signal_source.cc
+++ b/src/algorithms/signal_source/adapters/fifo_signal_source.cc
@@ -18,14 +18,12 @@
 #include "fifo_signal_source.h"
 #include "configuration_interface.h"
 #include "fifo_reader.h"
-#include "gnss_sdr_filesystem.h"
 #include "gnss_sdr_flags.h"
 #include "gnss_sdr_string_literals.h"
-#include "gnss_sdr_valve.h"
 #include <glog/logging.h>
+#include <gnuradio/blocks/file_sink.h>
+#include <gnuradio/blocks/file_source.h>
 #include <cmath>  // ceil, floor
-#include <fstream>
-#include <utility>  // move
 
 
 using namespace std::string_literals;
--- a/src/algorithms/signal_source/adapters/fifo_signal_source.cc
+++ b/src/algorithms/signal_source/adapters/fifo_signal_source.cc
@@ -28,7 +28,7 @@ using namespace std::string_literals;
 
 FifoSignalSource::FifoSignalSource(ConfigurationInterface const* configuration,
     std::string const& role, unsigned int in_streams, unsigned int out_streams,
-    Concurrent_Queue<pmt::pmt_t>* queue)
+    [[maybe_unused]] Concurrent_Queue<pmt::pmt_t>* queue)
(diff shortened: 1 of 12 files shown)


lenhart:
Hi, 
I don't know what the issue is with the failing `volk-gnsssdr-windows` test is. I cannot find references to any files I touched in the test's output...

Things which might stand out in my code style-wise:
- used a templated function (which I have seen elsewhere but not that often)
- just noticed that I used the c++14 language feature of digit separation in the udp source to make the fifo buffer size more intuitive (as in int foo = 1'000'000;)

I also just noticed that the cast I changed in the udp source was not a c-style cast, but I think a constructor `auto foo = size_t(some_uint64_variable)`. Shall I change that back? don't know what is more readable/performant. Both should work..

Thanks

src/algorithms/signal_source/adapters/fifo_signal_source.cc
```diff
@@ -0,0 +1,91 @@
+/*!
+ * \file file_source_base.cc
```
jwmelto, line 2: match file name

src/algorithms/signal_source/adapters/fifo_signal_source.h
```diff
@@ -0,0 +1,79 @@
+/*!
+ * \file file_source_base.h
```
jwmelto, line 2: match file name

src/algorithms/signal_source/gnuradio_blocks/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ endif()
 
 
 set(SIGNAL_SOURCE_GR_BLOCKS_SOURCES
+    fifo_reader.cc
```
jwmelto, line 17: does this need to be conditional on being a Unix build? I don't know enough about Windows, but you say this is for Unix FIFOs

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * -----------------------------------------------------------------------------
+ */
+
+#include "fifo_reader.h"
+#include <glog/logging.h>
+#include <cstdlib>
+#include <fstream>
+#include <iostream>
+#include <string>
```
jwmelto, line 23: review your includes and remove the ones not needed. For example, `<fstream>` is probably not needed here, it must be in the class header. `<iostream>` is only needed for the standard streams; prefer glog to cout. That said, there is plenty of cout littered through the code base, so follow the pattern

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -0,0 +1,111 @@
...
+
+// private constructor called by ::make
+FifoReader::FifoReader(const std::string &file_name, const std::string &sample_type)
+    : gr::sync_block("fifo_reader",
+          gr::io_signature::make(0, 0, 0),                    // no input
+          gr::io_signature::make(1, 1, sizeof(gr_complex))),  // <+MIN_OUT+>, <+MAX_OUT+>, sizeof(<+OTYPE+>)
+      file_name_(file_name),
+      sample_type_(sample_type)
+{
+    std::cout << "Starting FifoReader\n";
```
jwmelto, line 39: be sure to flush your std::cout stream ... don't assume it is a terminal and line-buffered.

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -0,0 +1,111 @@
...
+    // instream: https://en.cppreference.com/w/cpp/io/basic_ifstream
+}
+
+bool FifoReader::start()
+{
+    fifo_.open(file_name_, std::ios::binary);
+    if (!fifo_.is_open())
+        {
+            DLOG(ERROR) << "Error opening fifo\n";
```
jwmelto, line 48: DLOG is compiled out for non-debug builds. I'd expect any ERROR to be LOG.

Trailing newlines are not needed for glog streaming

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -0,0 +1,111 @@
...
+    return this->WORK_CALLED_PRODUCE;
+}
+
+// read gr_complex items from fifo
+size_t FifoReader::read_gr_complex(int noutput_items, gr_vector_void_star &output_items)
+{
+    size_t items_retrieved = 0;
+    for (int n = 0; n < noutput_items; n++)
+        {
+            std::vector<char> buffer(4);
```
jwmelto, line 94: this part is a bit obtuse. The vector is better than a C-array, but `std::array` would accomplish the same purpose, for fixed size reads. However, all you are really doing is reading into a complex, so that would make the most sense:
```c++
gr_complex sample;
fifo_.read((char*)&sample, sizeof(sample));
...
```

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -0,0 +1,111 @@
...
+            fifo_.read((char *)&buffer[0], buffer.size());
+            if (fifo_.good())
+                {
+                    gr_complex sample;
+                    memcpy(&sample, &buffer[0], sizeof(sample));
+                    static_cast<gr_complex *>(output_items.at(0))[n] = sample;
+                    items_retrieved++;
+                }
+            else if (fifo_.eof() || fifo_.fail())
```
jwmelto, line 103: EOF is one thing, fail is another. Or is the FIFO guaranteed to block? Note, blocking isn't a good fit for the GNU Radio paradigm, as I understand it, but I could be wrong.

jwmelto (commented):
This is a very nice PR. Your commits are atomic and each have clear purpose. I have no authority, but I like it.

lenhart:
> This is a very nice PR. Your commits are atomic and each have clear purpose. I have no authority, but I like it.

Thank you so much for the nice words and all the feedback :+1: . I'll fix it and reply to a couple of comments directly!

src/algorithms/signal_source/gnuradio_blocks/CMakeLists.txt
lenhart: It should compile without problems on windows as the c++ code does not contain any OS specific components. 
I have however neither knowledge nor a Windows machine around to test how FIFOs work under Windows.. My research did not leave me confident enough to say if it would work or not. Therefore I documented that I is implemented for a Unix FIFO to prevent confusion. Might work under Windows, but might also require some adaptation. I think it should be safe however to include it in the CMakeLists.txt
(or do you think that could possibly the reason why the windows volkgnsssdr build is failling?)

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
lenhart: I'll change it to glog as it is the only std::cout in this file. thanks!
(btw, is there a preference to use LOG() or DLOG() ? one is just defined as alias for the other..)

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
lenhart: changed vector to array in the templated read_interleave function in the header and changed this fct to read into gr_complex directly.
I started with a vector because I thought I'd read larger chunks, but then went for the easy/readable version instead.

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
lenhart: I agree that the handling of flags was wrong here. I removed .fail() from the case and added a 'final' else branch giving out errors for unhandled cases. I tried to provoke some by letting the FIFO run dry/feed in at a very slow pace, but did not succeed in getting that error. Guess this works alright (don't know how the situation would be with unbuffered writes to the fifo..)

about the blocking: I don't think that the read() will block. The open() does and caused problems when I tried to do it in the constructor. Therefore I placed it in the open() method which is meant for io drivers, etc. By doing so gnss-sdr initialized completely without problems.

carlesfernandez: Great! This was my only big concern on this PR, but it seems you already solved it.

lenhart:
@carlesfernandez 
i just noticed that the last commit broke a couple test which were succeeding before. 
Seems that changing cout to glog misses a header beside the general glog include.. 

Will fix that briefly; just don't accept the PR yet. Sorry that I did not check the result earlier, forgot it a bit because some take so long..

carlesfernandez:
I'm joining @jwmelto  in praising this PR 😄 

The volk-gnsssdr-windows test started to fail some days ago, without any code change from our side, so I guess the environment in GitHub CI system has changed something that breaks compilation. Anyway, it's nothing related to this PR.

However, the building is now failing since `fifo_reader.h` does not see `glog/logging.h`. We try to keep glog in the private part of the implementation, so the library does not need to expose the logging interface in public. However, you are using a template here, so it makes sense here to do that. Could you please fix this?

lenhart:
> I'm joining @jwmelto in praising this PR smile
> 
> The volk-gnsssdr-windows test started to fail some days ago, without any code change from our side, so I guess the environment in GitHub CI system has changed something that breaks compilation. Anyway, it's nothing related to this PR.
> 
> However, the building is now failing since `fifo_reader.h` does not see `glog/logging.h`. We try to keep glog in the private part of the implementation, so the library does not need to expose the logging interface in public. However, you are using a template here, so it makes sense here to do that. Could you please fix this?

Hi, thank you a lot too :-) 

The obvious fix would be to move the include into the header; before I push that I wanted to briefly discuss alternatives if you think any of these is preferable:
1. put the error output into a common private function which moves the one problematic line into the source file again (easy fix)
2. find a way to forward declare the log output (don't know if possible because as it is not accessed as a class member..)
3. define a library header only file for the type parsing:
    - pro: could be extended and used by other classes. (adding specific classes for other sources. Wouldn't have to reinvent the wheel)
    - pro: fct would become a pure functional function, which is always nice for threaded applications
    - con: I remember that passing streams around was a bit tedious
    - con: was there something problem with cmake and header only classes? I think I ran into that error once but am no expert in cmake..
    - con: one more file

Any preference? :-)

carlesfernandez:
The simplest should work here, so I vote for solution 1

carlesfernandez:
One more thing: We need your commits to be signed (see https://gnss-sdr.org/docs/tutorials/using-git/#sign-your-commits) as an indication of [Developer’s Certificate of Origin](https://github.com/gnss-sdr/gnss-sdr/blob/next/.github/DCO.txt) fulfillment.

Signing your last commit should be ok.

lenhart:
Ah yes, sorry about the signing. I remember that I turned that off some time ago because my password for the private key was not accepted when I used it with git (but worked elsewhere). I changed git to my other mail address and pgp key now, hope that doesn't bring in new problems elsewhere.

I did a local make clean and rebuild it to not have another include surprise, moved the problematic output into the source file, so the logging #include remains there.
Also added a [[maybe_unused]] to one of the constructor arguments to fix the only remaining warning (from my files).

cheers!

lenhart:
Oh. I resolved the merge conflict here on github web.. which seems to have merged next into my feature branch (noticed when I wanted to rebase my local branch to next).
Is this a problem? I have no idea to restore that to a clean state. What happens if I undo the merge and `push --force` the rebased branch? Do I want to find out? :D

lenhart:
oh, sorry about that, it is undone. What about `[[maybe_unused]]` ? Found that used elsewhere and used it in one of the later commits.

jwmelto:
> oh, sorry about that, it is undone. What about `[[maybe_unused]]` ? Found that used elsewhere and used it in one of the later commits.

Yes, the `[[maybe_unused]]` is C++17 (darn). That needs to go. And I'm likely the culprit who introduced it.

carlesfernandez:
All building tests are now passing, so I'm merging this. Thanks again for your work!

carlesfernandez:
Please note that ```[[maybe_unused]]``` seems to be happily ignored in C++11, and the compiler (at least GCC) doesn't complain about it, so I leave it there.

lenhart:
Hi! I am very happy that it worked out :-) Thanks both of you for the feedback! 

Good news about the [[maybe_unused]], think that is looks nicer and is more usable than the alternative `foo(type /*commented-name*/)` which also suppress the warning.

Cheers!

lenhart:
Ah, when I have a bit time to spare, shall I write a word or two about it (and other available signal sources) on the [signal source documentation page](https://gnss-sdr.org/docs/sp-blocks/signal-source/)?

PR #514, opened 2021-06-07
GSoC 21 - Trial pull request

harshadms:
Issuing a trial pull request. Added a comment in pcps_acquisition.cc

### DIFF
--- a/data/.gitignore
+++ b/data/.gitignore
@@ -4,3 +4,4 @@
 *
 # Except this file
 !.gitignore
+.bin
--- a/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
+++ b/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
@@ -12,7 +12,7 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  *
  * -----------------------------------------------------------------------------
- */
+*/
 
 #include "rtklib_pvt_gs.h"
 #include "MATH_CONSTANTS.h"
@@ -93,17 +93,20 @@ namespace bc = boost::integer;
 
 rtklib_pvt_gs_sptr rtklib_make_pvt_gs(uint32_t nchannels,
     const Pvt_Conf& conf_,
-    const rtk_t& rtk)
+    const rtk_t& rtk,
+    SpoofingDetector spoofing_detector)
 {
     return rtklib_pvt_gs_sptr(new rtklib_pvt_gs(nchannels,
         conf_,
-        rtk));
+        rtk,
+        spoofing_detector));
 }
 
 
 rtklib_pvt_gs::rtklib_pvt_gs(uint32_t nchannels,
     const Pvt_Conf& conf_,
-    const rtk_t& rtk) : gr::sync_block("rtklib_pvt_gs",
+    const rtk_t& rtk,
+    SpoofingDetector spoofing_detector) : gr::sync_block("rtklib_pvt_gs",
                             gr::io_signature::make(nchannels, nchannels, sizeof(Gnss_Synchro)),
                             gr::io_signature::make(0, 0, 0))
 {
@@ -136,7 +139,11 @@ rtklib_pvt_gs::rtklib_pvt_gs(uint32_t nchannels,
     d_dump = conf_.dump;
     d_dump_mat = conf_.dump_mat and d_dump;
     d_dump_filename = conf_.dump_filename;
+    
+    d_total_pvt_measurements = 0;
+
     std::string dump_ls_pvt_filename = conf_.dump_filename;
+    
     if (d_dump)
         {
             std::string dump_path;
@@ -174,7 +181,7 @@ rtklib_pvt_gs::rtklib_pvt_gs(uint32_t nchannels,
 
     d_type_of_rx = conf_.type_of_receiver;
 
-    // GPS Ephemeris data message port in
+    // GPS Ephemeris data message pspoofing_detectorort in
     this->message_port_register_in(pmt::mp("telemetry"));
     this->set_msg_handler(pmt::mp("telemetry"),
 #if HAS_GENERIC_LAMBDA
@@ -511,11 +518,14 @@ rtklib_pvt_gs::rtklib_pvt_gs(uint32_t nchannels,
     d_beidou_dnav_almanac_sptr_type_hash_code = typeid(std::shared_ptr<Beidou_Dnav_Almanac>).hash_code();
 
     d_start = std::chrono::system_clock::now();
+
+    d_spoofing_detector = spoofing_detector;
 }
 
 
 rtklib_pvt_gs::~rtklib_pvt_gs()
 {
+    std::cout << "\nTotal iterations: " << d_total_pvt_measurements << "\n";
     DLOG(INFO) << "PVT block destructor called.";
     if (d_sysv_msqid != -1)
         {
@@ -1813,8 +1823,11 @@ void rtklib_pvt_gs::initialize_and_apply_carrier_phase_offset()
 int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_items,
     gr_vector_void_star& output_items __attribute__((unused)))
 {
+    std::cout << "\nNumber of records in iteration: " << noutput_items;
+    
     for (int32_t epoch = 0; epoch < noutput_items; epoch++)
-        {
+        {   
+            d_total_pvt_measurements++;
             bool flag_display_pvt = false;
             bool flag_compute_pvt_output = false;
             bool flag_write_RTCM_1019_output = false;
@@ -1825,6 +1838,7 @@ int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_item
 
             d_gnss_observables_map.clear();
             const auto** in = reinterpret_cast<const Gnss_Synchro**>(&input_items[0]);  // Get the input buffer pointer
+            d_spoofing_detector.check_position_consistency(1, 1, 1, in);
             // ############ 1. READ PSEUDORANGES ####
             for (uint32_t i = 0; i < d_nchannels; i++)
                 {
@@ -2238,6 +2252,6 @@ int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_item
                         }
                 }
         }
-
+        
     return noutput_items;
 }
--- a/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.h
+++ b/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.h
@@ -34,6 +34,7 @@
 #include <string>                 // for string
 #include <sys/types.h>            // for key_t
 #include <vector>                 // for vector
+#include "spoofing_detector.h"
 
 /** \addtogroup PVT
  * \{ */
@@ -64,7 +65,8 @@ using rtklib_pvt_gs_sptr = gnss_shared_ptr<rtklib_pvt_gs>;
 
 rtklib_pvt_gs_sptr rtklib_make_pvt_gs(uint32_t nchannels,
     const Pvt_Conf& conf_,
-    const rtk_t& rtk);
+    const rtk_t& rtk,
+    SpoofingDetector spoofing_detector);
 
 /*!
  * \brief This class implements a block that computes the PVT solution using the RTKLIB integrated library
@@ -125,11 +127,13 @@ public:
 private:
     friend rtklib_pvt_gs_sptr rtklib_make_pvt_gs(uint32_t nchannels,
         const Pvt_Conf& conf_,
-        const rtk_t& rtk);
+        const rtk_t& rtk,
+        SpoofingDetector spoofing_detector);
 
     rtklib_pvt_gs(uint32_t nchannels,
         const Pvt_Conf& conf_,
-        const rtk_t& rtk);
+        const rtk_t& rtk,
+        SpoofingDetector spoofing_detector);
 
     void msg_handler_telemetry(const pmt::pmt_t& msg);
 
@@ -248,7 +252,8 @@ private:
     int32_t d_display_rate_ms;
     int32_t d_report_rate_ms;
     int32_t d_max_obs_block_rx_clock_offset_ms;
-
+    int32_t d_total_pvt_measurements;
+    
     uint32_t d_nchannels;
     uint32_t d_type_of_rx;
 
@@ -267,6 +272,8 @@ private:
     bool d_show_local_time_zone;
     bool d_waiting_obs_block_rx_clock_offset_correction_msg;
     bool d_enable_rx_clock_correction;
+
+    SpoofingDetector d_spoofing_detector;
 };
 
 
--- a/src/algorithms/libs/spoofing_detector.cc
+++ b/src/algorithms/libs/spoofing_detector.cc
@@ -1,0 +1,67 @@
+/*!
+ * \file spoofing_detector.cc
+ * \brief Library of anti-spoofing functions
+ * \author Harshad Sathaye sathaye.h(at)northeastern.edu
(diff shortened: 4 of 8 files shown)


src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -93,17 +93,20 @@ namespace bc = boost::integer;
...
+        rtk,
+        spoofing_detector));
 }
 
 
 rtklib_pvt_gs::rtklib_pvt_gs(uint32_t nchannels,
     const Pvt_Conf& conf_,
-    const rtk_t& rtk) : gr::sync_block("rtklib_pvt_gs",
+    const rtk_t& rtk,
+    SpoofingDetector spoofing_detector) : gr::sync_block("rtklib_pvt_gs",
```
jwmelto, line 109: and here? is the pass by value intentional?

You end up with 4 copies of this object created. First you allocated one on the heap to pass to the factory method (it was leaked). Then you made a copy of that by passing into the factory by value. Here you make a 3rd copy, by passing by value here. Finally, you make a 4th copy below. 

All these copies seem a bit pointless.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -174,7 +181,7 @@ rtklib_pvt_gs::rtklib_pvt_gs(uint32_t nchannels,
 
     d_type_of_rx = conf_.type_of_receiver;
 
-    // GPS Ephemeris data message port in
+    // GPS Ephemeris data message pspoofing_detectorort in
```
jwmelto, line 184: I hope this is just keyboard dribble, and not an intentional edit to the comment?

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -511,11 +518,14 @@ rtklib_pvt_gs::rtklib_pvt_gs(uint32_t nchannels,
     d_beidou_dnav_almanac_sptr_type_hash_code = typeid(std::shared_ptr<Beidou_Dnav_Almanac>).hash_code();
 
     d_start = std::chrono::system_clock::now();
+
+    d_spoofing_detector = spoofing_detector;
```
jwmelto, line 522: if you initialize this in the member initializer list, you at least have the opportunity to elide one of the copies.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -511,11 +518,14 @@ rtklib_pvt_gs::rtklib_pvt_gs(uint32_t nchannels,
...
 }
 
 
 rtklib_pvt_gs::~rtklib_pvt_gs()
 {
+    std::cout << "\nTotal iterations: " << d_total_pvt_measurements << "\n";
```
jwmelto, line 528: This kind of `std::cout` output is uninformative. There is no context to know what "Total iterations" relates to. A LOG message would be better.

But if you **must** use `std::cout`, flush the stream so the output is more-or-less in time order

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -1813,8 +1823,11 @@ void rtklib_pvt_gs::initialize_and_apply_carrier_phase_offset()
 int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_items,
     gr_vector_void_star& output_items __attribute__((unused)))
 {
+    std::cout << "\nNumber of records in iteration: " << noutput_items;
```
jwmelto, line 1826: did you mean to not newline terminate this output?

See above comment about `std::cout` in general and stream flushing.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -1813,8 +1823,11 @@ void rtklib_pvt_gs::initialize_and_apply_carrier_phase_offset()
...
+    
     for (int32_t epoch = 0; epoch < noutput_items; epoch++)
-        {
+        {   
+            d_total_pvt_measurements++;
```
jwmelto, line 1830: are they still teaching post-increment? That unfortunate habit dates back to the original K&R C book. Pre-increment is better, except in cases where you really need post-increment semantics.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -1825,6 +1838,7 @@ int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_item
 
             d_gnss_observables_map.clear();
             const auto** in = reinterpret_cast<const Gnss_Synchro**>(&input_items[0]);  // Get the input buffer pointer
+            d_spoofing_detector.check_position_consistency(1, 1, 1, in);
```
jwmelto, line 1841: What are the magic numbers `1, 1, 1,`?

Since this statement has no obvious effect here, you might add a comment explaining what this does.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -2238,6 +2252,6 @@ int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_item
                         }
                 }
         }
-
+
```
jwmelto, line 2255: I suspect that **clang-format** would squash this whitespace. Be sure to run it before your final commit.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.h
```diff
@@ -34,6 +34,7 @@
 #include <string>                 // for string
 #include <sys/types.h>            // for key_t
 #include <vector>                 // for vector
+#include "spoofing_detector.h"
```
jwmelto, line 37: ah. Here is the declaration. This needs to be included in **rtklib_pvt** (probably in the header, since you should make the detector a member), or else you should move the construction of the `SpoofingDetector` to this class.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.h
```diff
@@ -64,7 +65,8 @@ using rtklib_pvt_gs_sptr = gnss_shared_ptr<rtklib_pvt_gs>;
 
 rtklib_pvt_gs_sptr rtklib_make_pvt_gs(uint32_t nchannels,
     const Pvt_Conf& conf_,
-    const rtk_t& rtk);
+    const rtk_t& rtk,
+    SpoofingDetector spoofing_detector);
```
jwmelto, line 69: Consider that all you need to construct the detector is the config so skip this argument and just construct in the ctor.

Alternatively, follow the factory pattern in use everywhere else, and create a shared pointer that you pass in here.

src/algorithms/libs/spoofing_detector.cc
```diff
@@ -0,0 +1,67 @@
...
+ * Copyright (C) 2010-2021  (see AUTHORS file for a list of contributors)
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * -----------------------------------------------------------------------------
+ */
+
+#include "spoofing_detector.h"
+#include "GPS_L1_CA.h"
+#include <cmath>    // for floor, fmod, rint, ceil
+#include <cstring>  // for memcpy
```
jwmelto, line 20: `memcpy` is usually a security hole. Try to avoid it

src/algorithms/libs/spoofing_detector.cc
```diff
@@ -0,0 +1,67 @@
...
+
+SpoofingDetector::SpoofingDetector(const ConfigurationInterface* configuration)
+{
+    d_dump_results = configuration->property("Spoofing.dump_results", true);
+
+    // ####### Position consistency check configuration
+    d_position_check = configuration->property("Spoofing.position_check", false);
+    d_max_jump_distance = configuration->property("Spoofing.max_jump_distance", 100);
+    d_geo_fence_radius = configuration->property("Spoofing.geo_fence_radius", 15);
+    d_velocity_difference = configuration->property("Spoofing.geo_fence_radius", 15);
```
jwmelto, line 37: The interaction with the configuration more properly belongs in an adapter. Consider passing these values to the ctor rather than the configuration. This implementation hard-codes an unnecessary constraint that there can only be one "Spoofing" configuration. As implemented, this object is an element of the PVT, so this configuration would seem be more correctly be a sub-element of PVT.  Maybe it doesn't matter at this point, but it would be better to write more maintainable code than to assume that brittle code will "just be okay"

Use the member initializer list to initialize your members.

src/algorithms/libs/spoofing_detector.cc
```diff
@@ -0,0 +1,67 @@
...
+    DLOG(INFO) << "Spoofing detector destructor";
+}
+
+void SpoofingDetector::check_position_consistency(
+    double lat, 
+    double lon, 
+    double alt, 
+    const Gnss_Synchro **in)
+{
+    DLOG(INFO) << "Check position consistency function";
```
jwmelto, line 53: more detail here?

src/algorithms/libs/spoofing_detector.cc
```diff
@@ -0,0 +1,67 @@
...
+    // ####### Position consistency check configuration
+    d_position_check = configuration->property("Spoofing.position_check", false);
+    d_max_jump_distance = configuration->property("Spoofing.max_jump_distance", 100);
+    d_geo_fence_radius = configuration->property("Spoofing.geo_fence_radius", 15);
+    d_velocity_difference = configuration->property("Spoofing.geo_fence_radius", 15);
+
+    DLOG(INFO) << "Spoofing detector constructor";
+}
+
+SpoofingDetector::~SpoofingDetector()
```
jwmelto, line 42: there is no need to write what the compiler will happily generate for you. As it is, you should see several cases of this message in your output (unless you don't build a debug build, and DLOG is compiled out.

I recommend just eliding the do-nothing dtor declaration/definition.

src/algorithms/libs/spoofing_detector.cc
```diff
@@ -0,0 +1,67 @@
...
+    DLOG(INFO) << "Compare velocity function";
+}
+
+void SpoofingDetector::static_pos_check(
+    double lat, 
+    double lon, 
+    double alt)
+{
+    DLOG(INFO) << "Check static position function";
+}
```
jwmelto, line 67: Don't forget to run **clang-format** ... it would have added the trailing newline here for you (among other things)

src/algorithms/libs/spoofing_detector.h
```diff
@@ -0,0 +1,93 @@
...
+#include <fstream>
+#include <list>
+#include <map>
+#include <vector>
+#include <set>
+#include "gnss_synchro.h"
+#include <boost/circular_buffer.hpp>
+#include "configuration_interface.h"
+#include <glog/logging.h>
+#include <gflags/gflags.h>
```
jwmelto, line 30: don't include what you don't need. Try commenting out all of these includes and see what you actually need. For instance, you don't need `<iostream>` or `<fstream>`. You do need `<map>` but none of the other standard headers. You don't need ``<boost/circular_buffer.hpp>`, nor do you need the glog or gflags includes.

src/algorithms/libs/spoofing_detector.h
```diff
@@ -0,0 +1,93 @@
...
+    int agc_score;
+    int total_score;
+};
+
+struct pos_sol
+{
+    // ####### Structure to store position solution
+    double lat;
+    double lon;
+    double alt;
```
jwmelto, line 49: at a minimum, consider documenting your units. Meters? Feet? Furlongs?

src/algorithms/libs/spoofing_detector.h
```diff
@@ -0,0 +1,93 @@
...
+        // ####### General Functions
+        double distance(double lat1, double lon1, double lat2, double lon2);
+};
+
+#endif
+/*
+Position Jump - 1
+Compare Velocity - 2
+Static Position - 3
+*/
```
jwmelto, line 93: make sure all your files have a trailing newline. Run **clang-format**

data/.gitignore
```diff
@@ -4,3 +4,4 @@
 *
 # Except this file
 !.gitignore
+.bin
```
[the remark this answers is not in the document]
jwmelto: what do you think this change does? The existing file already ignored everything, so this change is pointless.

jwmelto (changes requested):
Hi @harshadms 

First, you should know that I carry absolutely no weight on this project, so if you are so inclined, you can completely ignore my comments. I don't recommend it, but it's an option.

Second, I would suggest that you create a branch off of your **next** and create the PR from there. Your **next** should always be in sync (or at least synchronizable) with this **next**.

Third, (or maybe first) be sure to read [CONTRIBUTING.md](https://github.com/gnss-sdr/gnss-sdr/blob/master/CONTRIBUTING.md). You will need to sign your commits (or at least, the last one) and there are a number of CI surprises awaiting you, but you can avoid some of the pitfalls by following the instructions.

Finally (in my opinion), a PR should be capable of being merged. This one does nothing. Merging this (in its current state) would be detrimental to the code base. Specifically, your SpoofingDetector class is a stub; it's not even a particularly well thought out stub. YMMV, but usually when I try to wire in an interface before I actually write the algorithm, I soon find out that my interface was incomplete, and I end up with more re-work than was necessary.

I'm much more interested in what your SpoofingDetector (eventually) DOES. Let's see your algorithms. What data do you bring in, and what output do you produce? What kind of decision might downstream processing make in response to the detection of a spoofer?

These data flows are not straightforward in GNSS-SDR (I'm also interested in the solution here), so they bear careful thought.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
harshadms: Noted, it was just for debugging, I have switched to dlog now.

carlesfernandez:
Hi @harshadms,

I assume this is the first PR trial belonging to the GSoC bonding period. In addition to all the comments by @jwmelto, which I totally agree with, you need to make sure that:

1) You have branched off from `next` (so, in a new branch created from the current `next`)
2) You have signed your commits.

Please check the documentation [here](https://gnss-sdr.org/docs/tutorials/using-git/#how-to-work-with-the-source-code-and-submit-your-changes).

harshadms:
Hi @carlesfernandez and @jwmelto. Yes, it was my first pull request.
@jwmelto thanks a lot for your suggestions, they are helpful. I'll incorporate these moving forward.
I have branched off from next and all my new commits will be on "gsoc_21_sd".
I have also reconfigured the git client to use gpg keys correctly; earlier there was an issue with the client. It won't be a problem for future commits.
 
Sorry for the mess; it's my first time contributing to an open-source project. Still getting the hang of things.

carlesfernandez:
Hi @harshadms,

don't worry. We ask students for this trial PR in order to make them follow the workflow at least once, before any actual contribution, so they can get acquainted with the methodology and tools set up. Basically, the workflow is as follows:

1. Go to the current next branch in your local repository:
```
git checkout next
```
2. Get the updates from the upstream repo:
```
git pull upstream next
```
3. Create a new branch:
```
git checkout -b my-branch
```
4. Work in your new branch. Remember to sign your commits!

5. Push to your repo:
```
git push origin my-branch
```
6. When ready, do the Pull-Request (button on GitHub page).

7) Once your PR is accepted, you can delete your local branch and start over again:
```
git checkout next
git branch -D my-branch
git pull upstream next
git checkout -b my-other-branch
...
```

Issue #515, opened 2021-06-09
Two_Bit_Packed_File_Signal_Source doesn't output the correct data type


Rigel09:
When using ```Two_Bit_Packed_File_Signal_Source```  the output never seems to match the input that ```Signal Conditioner``` is expecting. I get an item mismatch between signal source and signal conditioner.

```
Can't connect SignalSource to SignalConditioner : itemsize mismatch: valve0:0 using 8, interleaved_char_to_complex0:0 using 1
Unable to connect flowgraph
```

Here are my settings for the conf file. 
```
;######### SIGNAL_SOURCE CONFIG ############
SignalSource.implementation=Two_Bit_Packed_File_Signal_Source
SignalSource.filename=/Path/To/File
SignalSource.item_type=byte
SignalSource.sampling_frequency=60000000
SignalSource.sample_type=iq

;######### SIGNAL_CONDITIONER CONFIG ############
SignalConditioner.implementation=Signal_Conditioner
DataTypeAdapter.implementation=Ibyte_To_Complex
InputFilter.implementation=Pass_Through
Resampler.implementation=Pass_Through
```

Any help is appreciated.

jwmelto:
It's entirely possible that I broke this. The two-bit file sources are interesting. 

- There is `.item_type` that drives the input. This defaults to "byte". This defines the size of a single input item. The only supported values are "byte" and "short" (although I have my doubts about "short" in little-endian architectures)
- There is `.sample_type` that defines whether the input is raw amplitude ("real"), or I/Q ("iq" or "qi")
- There is also `.big_endian_items` (default to `true`) that affects how "short" items are read. I think it tells if the data is stored in Big Endian form (network order) or not.
- There is also `.big_endian_bytes` (default to `false`). This flag and the prior one are passed to the `unpack_2bit_samples` block constructor

The 2-bit file source inserts an "unpack" block followed by a "char to float" block between the file reader and the signal conditioner. So the output of this file source is either float or complex. Since you specified "iq", it will be complex.

It looks like your signal conditioner is superfluous; the samples are already complex coming out of the signal source. What if you just make your SignalConditioner `Pass_Through`?

Rigel09:
Thanks for the reply.

I gave that a shot and it failed as well. Here is the error it gave.

'''
[...]

PR #517, opened 2021-06-11
return the correct right block

jwmelto:

[comment]: # (
)

[comment]: # (
SPDX-FileCopyrightText: 2011-2020 Carles Fernandez-Prades <carles.fernandez@cttc.es>
)

This PR addresses the problem #515 where a two-bit packed file source cannot be configured into the graph, because it incorrectly returns the wrong right-hand block. This results in a size mis-match between the signal source and the signal conditioner.

@carlesfernandez don't merge this too quickly; I haven't had time to look for other signal sources that might suffer from the same problem.


jwmelto:
I don't think this fixes anything. All the signal sources have a virtual `source()` method that returns the rightmost block. The base class `get_right_block()` method calls it.

All I did here was short-circuit that re-direction.

PR #518, opened 2021-06-12
change sanity check to check the right thing

jwmelto:
[comment]: # (
)

[comment]: # (
SPDX-FileCopyrightText: 2011-2020 Carles Fernandez-Prades <carles.fernandez@cttc.es>
)

This PR **correctly** (?!) addresses issue #515 by ensuring that the sanity check done in **gnss_flowgraph** tests the right thing.

The issue is highlighted by the **two_bit_packed_file_source** that _reads_ a byte but produces multiple `gr_complex` samples per byte. The sanity check test compared the **input** item size with the expected input size of the Signal Conditioner.

The fix is to explicitly extract the **output** item size from the block.

One might argue that the validity checks done here don't need to be done; the blocks won't connect if the output/input don't match. However, as a bug fix, this minimal change is submitted.


carlesfernandez:
Thanks! Merging

PR #526, opened 2021-07-09
PVT solution consistency checks

harshadms:
A collection of functions to ensure consistency of obtained PVT solution to detect GNSS spoofing attacks.

These include:

1. Position jumps
2. Velocity consistency
3. Comparison with a static surveyed position
4. Abnormal PVT solution checks
5. Overshadow attack detection using correlator output

Consistency checks 1 through 4 are implemented in the PVT block and check #5 is implemented in the tracking block 
A score-based system is implemented to indicate the reliability of the calculated PVT solution.


jwmelto (changes requested):
I need to start with my standard disclaimer that I have no authority on this project, so feel free to disregard my comments if you must.

Since this is a GSoC project, I'll add to that that you may not be used to writing production-quality code, and you've probably never had as detail-oriented a reviewer as me look at your code before. Please be aware that comments on the code are not comments on you. Finally, your new/modified code is evaluated on its own merits. You may completely follow an existing pattern; since I'm not looking at the pattern and only your mods,  poor C++ practices will be criticized. Those may be the kinds of comments you choose to ignore, or at least file away for future reference. 

This is an ambitious and far-reaching PR, touching a lot of files. I would have liked to seen smaller chunks; because there is so much, and it all builds on some basic decisions, it is hard to address the issues. While there are numerous code quality issues, focusing on them is the wrong priority when there are design concerns. What I find lacking is an overarching vision of how your tests detect spoofing. I see various counters, but it's not obvious what phenomena you are detecting. 

Spoofing detection should be rooted in physics that the spoofer can't override. Comments are your friend. It would be extremely useful for each of your tests to describe **what** they are testing, **why** it's an indication of spoofing, and **how** the test is designed. As noted in the position-related tests, it appears that there is an assumption that the receiver isn't moving. That's not likely to be a useful check, since a stationary receiver should be considered an edge case.

I would encourage you to read anything from [Scott Meyers](https://www.aristeia.com) that you can. Another priceless resource is [Herb Sutter](https://herbsutter.com/elements-of-modern-c-style/). Have friends read your code and tell you what they think it does. The ability of code to be read is a key factor in maintainability. Especially in an open-source project, where the next person to touch your code may be continents away from you, it is important that your code clearly communicates your intent, both to the reader and to the computer. Towards that end, meaningful variable names, and functions that are of the form _verbNoun_ are important, with returns that match the intuition.

For example, `void position_jump()` implies that it will "jump" the position (whatever that means). Even if it were `check_position_jump()`, one would expect a "check" function to return the results of the check. Perhaps a more natural construct would be `bool didPositionJump()`, which would return true if the position jump is detected. 

This doesn't feel like "ready to ship" code; it reads more like a homework assignment. You can improve the quality by refining the design, further isolating your changes from the existing PVT and by more clearly defining precisely what it is you are testing for, and why that's a meaningful test. See `abnormal_position_checks()`; it computes a score that appears largely random. 

Don't forget to use the tools available to check your code, like **clang-format** and **clang-tidy**. They will get run before this can be merged anyway, so you might as well run them now and fix what they find.

carlesfernandez:
@harshadms  Please [sign](https://gnss-sdr.org/docs/tutorials/using-git/#sign-your-commits) your following commits when addressing @jwmelto comments.

Issue #485, opened 2021-03-09
Monitor Decimation Factor Not Working Properly


jt3232:
Hi, 

I was wanting to slow the number of messages coming from the monitoring. So I changed the decimation factor (Monitor.decimation_factor) from 1 to 50 as shown in the documentation examples. I then noticed that I was not getting messages every second. In fact, I didn't receive any for a 5 minute period that I waited. I changed the dec factor to 2 and then I got messages every ~20 seconds. I have 32 channels, so when decfac=1, I get 1600 messages every second. With dec fac=2, I get 32 messages every ~20 seconds.

I am receiving these messages using LabVIEW. 

Also, I don't receive any value for carrier_phase_rads as shown to be ID #15 in the gnss_synchro.proto file. I need this one :(

Thanks guys, you all rock!

Jaxon

jt3232:
The decimation factor still does not seem to work properly. It looks like it is used in gnss_synchro_monitor.cc in the general_work() function. I am figuring that in that loop, it checks for how many items are in that stream for each channel as it loops through each channel. Assuming one sets decmation factor to 2, it would send data if the channel has more than one item. However, if the channel only has 1 item most of the time, it will skip over the channel and not send anything, and when it finishes the loop, it wipes the items channel. Thus, if every time it checks the number of items in the channel and it only contains one item, it will never send the data over UDP. However, very rarely, if the decimation is set to 2, it will send them. I'm talking once every minute per channel on average. If decimation is 1, it sends 50 messages per channel per second, significantly faster than decimation 2 (which should give 25 messages per channel per second).

Thus, I think there may be some sort of bug in this loop. I am not a coding expert and I only barely understand C++/GNSS-SDR, so I hesitate to attempt to make any changes.

jt3232:
I am learning to code C++, so I printed. the length of each channel and they are indeed 1. Thus, my theory is true. The code has an issue. 

I do not know how exactly this can be changed, perhaps by putting the consume in with the if statement where it prints out? I do not know. I will try some more stuff out.

jt3232:
Placing the consume inside the if statement asking if the decimation factor has been reached has appeared to make the code function properly. I see the channels sending a message every 1 second with decimation set to 50, the proper functionality. I will look into creating a pull request.

PR #530, opened 2021-07-20
Fixed GNSS Synchro Monitor Decimation

jt3232:
The GNSS Synchro Monitor Decimation does not currently work properly. If the decimation value is set to 50, then one should receive one message per channel every second. However, this is not the case. When set to 50, the GNSS Synchro messages are never sent (on my system). I made an issue on the GitHub shedding more details on this problem. #485 

I have debugged this and determined this line to be the cause of the issue. Instead of being after the loop where it is ran every time, it needs to only be run when the decimation value is reached, in the if statement. Preliminary tests show this seems to work properly. 

The only thing I cannot confirm is that when the if statement is reached, is the most recent or oldest of the GNSS Synchro messages sent? I would like to believe that it is the most recent GNSS Synchro message that is sent since it is at the end of the vector. I cannot confirm this, so I hope someone else out there may be able to confirm this for me.

Jaxon

### DIFF
--- a/src/core/monitor/gnss_synchro_monitor.cc
+++ b/src/core/monitor/gnss_synchro_monitor.cc
@@ -84,10 +84,10 @@ int gnss_synchro_monitor::general_work(int noutput_items __attribute__((unused))
                             udp_sink_ptr->write_gnss_synchro(stocks);
                             // Reset count variable
                             count = 0;
+                            // Consume the number of items for the input stream channel
+                            consume(channel_index, ninput_items[channel_index]);
                         }
                 }
-            // Consume the number of items for the input stream channel
-            consume(channel_index, ninput_items[channel_index]);
         }
 
     // Not producing any outputs


carlesfernandez:
Hi @jt3232 

Thanks for the fix. It seems to me that the most recent GNSS Synchro is sent since we are writing `in[channel_index][item_index]`, where `item_index` is the index of the object that fulfills the condition (that is, the latest).

Could I ask you to [sign your commit](https://gnss-sdr.org/docs/tutorials/using-git/#sign-your-commits)?

jt3232:
Wow! I read the guidelines but must have skipped over that part, my apologies.

Here is what I did. I got on my branch and did the following commands:

`git commit --amend --signoff`

I left everything in the file as it was, I just exited the Vim editor.
Then I ran

`git push origin -f <My Branch Name>`

I hope that solved the issue.

Thanks,

Jaxon

carlesfernandez:
Thanks! Merged.

PR #539, opened 2021-08-24
change the packets per sample calculation for 2-bit files.

jwmelto:
See #518 
This is complicated and I botched it before.
Hopefully someone with access to 2-bit samples can test this


carlesfernandez:
Thanks! Merging into next.

If someone can share a test file for this format, it would be great.

PR #549, opened 2021-09-21
added 180 deg. inverted carrier phase PLL lock flag in gnss_synchro

mmajoral:
added 180 deg. inverted carrier phase PLL lock flag in gnss_synchro


jwmelto:
This change just reports the inverted phase? No logic uses that information?

carlesfernandez:
I’m always hesitant to include new parameters to gnss_synchro.h and the corresponding .proto file, although ProtocolBuffers allows doing so without breaking existing implementations based on the former .proto file. Furthermore, in this case, the new parameter is never used downstream in the receiver chain.

However, there are interesting use cases here, for instance gnss-sdr feeding a signal rebroadcaster. In such a case, this flag would be needed for disambiguating the received data phase. If no better mechanism is found, we will merge this PR.

PR #552, opened 2021-09-23
document PVT params; enable RTKLIB trace level to be set

jwmelto:
This PR provides the ability to configure the RTKLIB trace level. Without this, an error might look like this:
```
RTKLIB rtkpos error: 12:29:39.16: point pos error ()
```
which is not very informative.

If the trace level is set to something > 2, then instead you might see
```
RTKLIB rtkpos error: 12:29:39.16: point pos error (chi-square error nv=8 vv=243.1 cs=10.8)
```

This adds a new attribute to the PVT block, `.rtk_trace_level`, which defaults to 0 (off).

----------------------
One side note: I'm working on a far-reaching change to ensure that all output to stdout or stderr is flushed explicitly, instead of relying on the behavior of TTYs to line-buffer. I made that change in the files I modified to simplify my merging later. It is unrelated to this change, but useful on its own.


carlesfernandez:
I like this change! Entering in the testing pipeline...

carlesfernandez:
Hi @jwmelto,

For consistency, I think `Pvt_Conf` should be kept as a class in pvt_conf.h:

```
class Pvt_Conf
{
public:
    Pvt_Conf() = default;
 
    ...
};
```

Please note that it is forward-declared as a class at https://github.com/gnss-sdr/gnss-sdr/blob/dcf9c8c7749d9cf119fbae9ed33e50588fa77d31/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.h#L57

jwmelto:
> For consistency, I think `Pvt_Conf` should be kept as a class in pvt_conf.h:

I don’t feel strongly about it. When I changed from `class` to `struct` I looked for a compile warning like I expected that forward declaration to cause and didn’t see it. 

FWIW there is really no difference between a `class` and a `struct` except default visibility. Explicitly declaring the ctor `default` has no effect in this case; eliding the declaration does the same thing.

carlesfernandez:
Well, I'm getting this warning with Clang:

```
In file included from /Users/carlesfernandez/git/cttc/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc:53:
/Users/carlesfernandez/git/cttc/src/algorithms/PVT/libs/pvt_conf.h:30:1: warning: 'Pvt_Conf' defined as a struct here but previously declared as a class; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
struct Pvt_Conf
^
/Users/carlesfernandez/git/cttc/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.h:57:1: note: did you mean struct here?
class Pvt_Conf;
^~~~~
struct
```

So maybe forward-declaring it as a struct would make everybody happy (although Clang recognizes it as a valid usage as it is)

jwmelto:
I can fix it however you like, but it won’t be until tomorrow

carlesfernandez:
Thanks!

I've added an item to the CHANGELOG and documented the new parameter in the [website](https://gnss-sdr.org/docs/sp-blocks/pvt/).

PR #557, opened 2021-10-06
Fixed default value of coherent_integration_time_ms property

iucgnaw:
Signed-off-by: iucgnaw <iucgnaw@msn.com>

[comment]: # (
)

[comment]: # (
SPDX-FileCopyrightText: 2011-2021 Carles Fernandez-Prades <carles.fernandez@cttc.es>
)

Any code contributions going into GNSS-SDR will become part of a GPL-licensed,
open source repository. It is therefore imperative that code submissions belong
to the authors, and that submitters have the authority to merge that code into
the public GNSS-SDR codebase.

For that purpose, we use the
[Developer's Certificate of Origin](https://github.com/gnss-sdr/gnss-sdr/blob/next/.github/DCO.txt).
It is the same document used by other projects. Signing the DCO states that
there are no legal reasons to not merge your code.

To sign the DCO, suffix your git commits with a `Signed-off-by:` line. When
using the command line, you can use `git commit -s` to automatically add this
line. If there were multiple authors of the code, or other types of
stakeholders, make sure that all are listed, each with a separate
`Signed-off-by:` line.

Before submitting your pull request, please make sure the following is done:

1.  You undertake the
    [Contributor Covenant Code of Conduct](https://github.com/gnss-sdr/gnss-sdr/blob/main/CODE_OF_CONDUCT.md).
2.  You have read the
    [Developer's Certificate of Origin](https://github.com/gnss-sdr/gnss-sdr/blob/next/.github/DCO.txt)
    and
    [signed your commits](https://gnss-sdr.org/docs/tutorials/using-git/#sign-your-commits)
    as an indication of fulfillment.
3.  You have read the
    [Contributing Guidelines](https://github.com/gnss-sdr/gnss-sdr/blob/main/CONTRIBUTING.md).
4.  You have read the [coding style guide](https://gnss-sdr.org/coding-style/).
    Specifically, you have read
    [about clang-format](https://gnss-sdr.org/coding-style/#use-tools-for-automated-code-formatting)
    and [about clang-tidy](https://gnss-sdr.org/coding-style/#use-code-linters),
    and you have applied those tools to your changes.
5.  You have forked the
    [gnss-sdr upstream repository](https://github.com/gnss-sdr/gnss-sdr) and
    have created your branch from `next` (or any other currently living branch
    in the upstream repository).
6.  Please include a description of your changes here.

**Please feel free to delete this line and the above text once you have read it
and in case you want to go on with your pull request, and explain your intend
below.**

### DIFF
--- a/src/algorithms/acquisition/libs/acq_conf.cc
+++ b/src/algorithms/acquisition/libs/acq_conf.cc
@@ -70,7 +70,7 @@ void Acq_Conf::SetFromConfiguration(const ConfigurationInterface *configuration,
     const int64_t fs_in_deprecated = configuration->property("GNSS-SDR.internal_fs_hz", fs_in);
     fs_in = configuration->property("GNSS-SDR.internal_fs_sps", fs_in_deprecated);
     doppler_max = configuration->property(role + ".doppler_max", doppler_max);
-    sampled_ms = configuration->property(role + ".coherent_integration_time_ms", sampled_ms);
+    sampled_ms = configuration->property(role + ".coherent_integration_time_ms", ms_per_code);
     bit_transition_flag = configuration->property(role + ".bit_transition_flag", bit_transition_flag);
     max_dwells = configuration->property(role + ".max_dwells", max_dwells);
     dump = configuration->property(role + ".dump", dump);


src/algorithms/acquisition/libs/acq_conf.cc
```diff
@@ -70,7 +70,7 @@ void Acq_Conf::SetFromConfiguration(const ConfigurationInterface *configuration,
     const int64_t fs_in_deprecated = configuration->property("GNSS-SDR.internal_fs_hz", fs_in);
     fs_in = configuration->property("GNSS-SDR.internal_fs_sps", fs_in_deprecated);
     doppler_max = configuration->property(role + ".doppler_max", doppler_max);
-    sampled_ms = configuration->property(role + ".coherent_integration_time_ms", sampled_ms);
+    sampled_ms = configuration->property(role + ".coherent_integration_time_ms", ms_per_code);
```
jwmelto, line 73: Can you explain the purpose behind this change? Both values default to 1, and there is no parameter to change `ms_per_code`, so on first reading, I can't see that this does anything?

carlesfernandez:
Hi all,

this change has no effect, since the default value is the same. Still, I can see the reason of this PR since, semantically, the name of the variable makes more sense here.

iucgnaw:
Normally we don't specify Acquisition_1B.coherent_integration_time_ms in conf file, the original code  will cause sampled_ms use default value 1.
If don't change this to ms_per_code, the default value 1 will always cause warning by following check:
    if ((sampled_ms % ms_per_code) != 0)
        {
            LOG(WARNING) << "Parameter coherent_integration_time_ms should be a multiple of "
                         << ms_per_code << ". Setting it to " << ms_per_code;
            sampled_ms = ms_per_code;
        }
I believe the warning is non-sense and should be fixed by use default ms_per_code, which also reflect the description of Acquisition_1B.coherent_integration_time_ms in documentation:
"Set the integration time T_int, in ms. Should be a multiple of 4 ms. It defaults to *4 ms*." not 1 ms.

iucgnaw:
Forget to mention, when run to the below code:
sampled_ms = configuration->property(role + ".coherent_integration_time_ms", ms_per_code);
ms_per_code has already been set to 4, not the default value 1 set in Acq_Conf::Acq_Conf(), which is correct for GALILEO E1B acquisition. So it is not as @jwmelto mentioned not doing anything.

jwmelto:
No offense. Your meaning was clear. Perhaps mine was not.

I always look at code in terms of readability and maintainability. When you have a class (struct), then all the operations on that data should be encapsulated in class methods. This is good software practice.

In this case, you have a member that has a default value. Your logic depends on that default value being overwritten by code that is not part of this class. I assume it is coming from a config file, although it could be directly. I didn't look at the whole code base, just at the file you changed.

No matter what, your change does not make things worse. I'm (at this point) only suggesting an opportunity to make it better. Here's my point:

In **galileo_e1_pcps_ambiguous_acquisition.cc** (for example), the adapter constructor has this code (that I think you are referring to:

```c++
    acq_parameters_.ms_per_code = 4;
    acq_parameters_.SetFromConfiguration(configuration_, role, GALILEO_E1_CODE_CHIP_RATE_CPS, GALILEO_E1_OPT_ACQ_FS_SPS);
```

The problem with this is that `SetFromConfiguration` is a kind of lazy constructor. It pulls values from the configuration and overwrites the appropriate structure members. As a caller of this method, you (should) have no knowledge of what configuration parameters are support and what are not. Therefore, it's entirely possible that `ms_per_code` could be specified by parameter (it isn't today, but I hope it's easy to see how such a change might happen in the future). Consequently, the value set here might easily be overwritten in the method call.

Also, the pattern,

```c++
    foo = configuration->property("some_property", foo);
```

is a common way to implement an optional parameter.

Your change replaces that paradigm with a slightly more complex one. Unnecessary complexity is bad.

The real problem that you are addressing was caused by the unencapsulated change to a single configuration element, resulting in the violation of an un-stated constraint (namely that the integration time be a multiple of the code duration). Possibly a more robust change would be to add a mutator for the code duration that would set both `ms_per_code` and `sampled_ms` together, to coherent values.

Does that make sense? You ran into a bug, and while your change will fix the bug, it won't prevent that kind of abuse from happening in the future. I'd rather see an improvement in encapsulation, resulting in more maintainable code.

But my opinion doesn't matter. It's just another perspective.

PR #584, opened 2022-02-03
Added FIFO ibyte support and code cleanups

stefanlinden:
This PR adds support for ibyte input to the FIFO signal_source block. The output is (still) converted to gr_complex.

Furthermore, an additional buffer (based on the one used in `gr_complex_ip_packet_source.cc`) improves performance of the FIFO pipe reading slightly. Do note that it could add latency to low sample rate configurations so this might need to be made configurable.

Finally, some code cleanup such as removing function definitions from the `fifo_reader` header file and applying `clang-format` and `clang-tidy`.

It is still not as fast as a file signal_source block and still some code is repeated between the 'interleaved' sample handling and the `gr_complex` sample handling, so future optimisations are definitely possible.

### DIFF
--- a/src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
+++ b/src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
@@ -17,7 +17,6 @@
 
 #include "fifo_reader.h"
 #include <glog/logging.h>
-#include <volk/volk.h>
 
 // initial construction; pass to private constructor
 FifoReader::sptr FifoReader::make(const std::string &file_name, const std::string &sample_type)
--- a/src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
+++ b/src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
@@ -17,6 +17,9 @@
 
 #include "fifo_reader.h"
 #include <glog/logging.h>
+#include <volk/volk.h>
+
+const int FIFO_SIZE = 1472000;  // Value taken from gr_complex_ip_packet_source, is it optimal?
 
 // initial construction; pass to private constructor
 FifoReader::sptr FifoReader::make(const std::string &file_name, const std::string &sample_type)
@@ -30,7 +33,10 @@ FifoReader::FifoReader(const std::string &file_name, const std::string &sample_t
           gr::io_signature::make(0, 0, 0),                    // no input
           gr::io_signature::make(1, 1, sizeof(gr_complex))),  // <+MIN_OUT+>, <+MAX_OUT+>, sizeof(<+OTYPE+>)
       file_name_(file_name),
-      sample_type_(sample_type)
+      sample_type_(sample_type),
+      fifo_buffer_(new char[FIFO_SIZE]),
+      buffer_idx_(0),
+      buffer_size_(0)
 {
     DLOG(INFO) << "Starting FifoReader";
 }
@@ -61,12 +67,14 @@ int FifoReader::work(int noutput_items,
     size_t items_retrieved = 0;
     if (sample_type_ == "ishort")
         {
-            // ishort == int16_t
             items_retrieved = read_interleaved<int16_t>(noutput_items, output_items);
+        }
+    else if (sample_type_ == "ibyte")  // Does this also work with cbyte?
+        {
+            items_retrieved = read_interleaved<int8_t>(noutput_items, output_items);
         }
     else if (sample_type_ == "gr_complex")
         {
-            LOG(WARNING) << sample_type_ << " is not yet tested. Please consider removing this warning if tested successfully";
             items_retrieved = read_gr_complex(noutput_items, output_items);
         }
     else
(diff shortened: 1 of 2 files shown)


src/algorithms/signal_source/gnuradio_blocks/fifo_reader.h
```diff
@@ -50,41 +49,16 @@ class FifoReader : virtual public gr::sync_block
...
+    //! \brief Destructor
+    //! private destructor
+    ~FifoReader();
+
     size_t read_gr_complex(int noutput_items, gr_vector_void_star &output_items);
+
     //! function to read data out of FIFO which is stored as interleaved I/Q stream.
     //! template argument determines sample_type
     template <typename Type>
-    size_t read_interleaved(int noutput_items, gr_vector_void_star &output_items)
```
jwmelto, line 57: Template definitions need to be visible at the point of instantiation. Most of the time, that means defining them in the header. It's possible that there is no harm, since this is a private method, but knowing that most developers aren't aware of these nuances, I prefer to comment at a minimum why it's safe to define this out of line.

An alternative to in-line definition is to move the definition after the class declaration in this file. That keeps the declaration easier to read, but satisfies the visibility requirements.

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.h
```diff
@@ -93,6 +67,9 @@ class FifoReader : virtual public gr::sync_block
     const std::string file_name_;
     const std::string sample_type_;
     std::ifstream fifo_;
+    char *fifo_buffer_;
```
jwmelto, line 70: don't use naked pointers. If you want dynamic sizing, use `std::vector<char>`. If you are ok with static sizing, use `std::array<char>`. Or, if you must, use `std::unique_ptr<char[]>` (see [SO](https://softwareengineering.stackexchange.com/questions/380176/should-i-use-a-unique-ptr-with-an-array-type-or-a-vector))

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.h
```diff
@@ -93,6 +67,9 @@ class FifoReader : virtual public gr::sync_block
...
+    int buffer_idx_, buffer_size_;
```
jwmelto, line 71: please declare each variable in a separate statement. Note that `buffer_size_` is useless information, if you use a container

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.h
jwmelto: or maybe you want a `boost::circular_buffer`?

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -30,7 +33,10 @@ FifoReader::FifoReader(const std::string &file_name, const std::string &sample_t
           gr::io_signature::make(0, 0, 0),                    // no input
           gr::io_signature::make(1, 1, sizeof(gr_complex))),  // <+MIN_OUT+>, <+MAX_OUT+>, sizeof(<+OTYPE+>)
       file_name_(file_name),
-      sample_type_(sample_type)
+      sample_type_(sample_type),
+      fifo_buffer_(new char[FIFO_SIZE]),
```
jwmelto, line 37: Just as a rule of thumb, if you are calling `new` explicitly, you've missed the modern C++ way. Either containers or `make_unique<T>`/`make_shared<T>` should be preferred.

Using RAII constructs has the (intentional) benefit of relieving you of destructor responsibilities.

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -61,12 +67,14 @@ int FifoReader::work(int noutput_items,
     size_t items_retrieved = 0;
     if (sample_type_ == "ishort")
         {
-            // ishort == int16_t
```
jwmelto, line 64: that's not a bad comment. Why remove it?

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -80,35 +88,102 @@ int FifoReader::work(int noutput_items,
...
+{
+    boost::mutex::scoped_lock lock(d_mutex);  // hold mutex for duration of this function
+    size_t items_retrieved = 0;
+    int n;
+    Type real;
+    Type imag;
+
+    if (buffer_idx_ >= buffer_size_)
+        {
+            fifo_.read(fifo_buffer_, FIFO_SIZE);
```
jwmelto, line 103: I realize this is moved code, so your responsibility is limited. However, I'd make a point of noting that this will block until FIFO_SIZE is read completely. At least, I think that's the design.

Now, the question is, should you read more than `noutput_items`, since that's all you wanted?

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -80,35 +88,102 @@ int FifoReader::work(int noutput_items,
...
+            else
+                {
+                    fifo_error_output();
+                    return 0;
+                }
+        }
+
+    for (n = 0; n < noutput_items && buffer_idx_ < buffer_size_; n++)
+        {
+            memcpy(&real, &fifo_buffer_[buffer_idx_], sizeof(real));
```
jwmelto, line 124: these copies can be replace with pointers:
```c++
auto real = reinterpret_cast<Type const*>(&fifo_buffer_[buffer_idx_]);
auto imag = reinterpret_cast<Type const*>(&fifo_buffer_[buffer_idx_ + sizeof(Type)]);
static_cast<gr_complex*>(output_items[0])[n] = gr_complex(*real, *imag);
```

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -80,35 +88,102 @@ int FifoReader::work(int noutput_items,
...
 
 void FifoReader::fifo_error_output() const
 {
     LOG(ERROR) << "unhandled FIFO event";
 }
+
+
+FifoReader::~FifoReader()
+{
+    delete[] fifo_buffer_;
```
jwmelto, line 188: prefer RAII to explicit memory management.

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
jwmelto: I do see the previous implementation only read `noutput_items` from the FIFO

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
```diff
@@ -80,35 +88,102 @@ int FifoReader::work(int noutput_items,
...
     size_t items_retrieved = 0;
-    for (int n = 0; n < noutput_items; n++)
+    int n;
+    gr_complex sample;
+
+    if (buffer_idx_ >= buffer_size_)
         {
-            gr_complex sample;
-            fifo_.read(reinterpret_cast<char *>(&sample), sizeof(sample));
+            fifo_.read(fifo_buffer_, FIFO_SIZE);
```
jwmelto, line 148: Note that the FIFO is buffered (`ifstream`) so you aren't gaining anything by buffering yourself here. If you have empirical evidence that reading one item at a time is "too slow" and that this buffering is much faster, then you owe future readers a comment to that effect. However, I'm hard pressed to see how this buffering -> single item extraction can be faster than the standard library.

jwmelto (commented):
It seems like this change was much more intrusive than simply supporting the `ibyte` data type. Perhaps break it into two PRs, one for `ibyte` and another for performance changes (if they are needed)?

Stylistic differences aside, my own experience with data sources was a difficult learning curve. One of the main points (that isn't sufficiently documented in GNU Radio) is that a source should block until it can produce the requested number of elements. It can return earlier, if desired, but isn't expected to. The code here blocks until an arbitrary number of bytes are read, which could have unpleasant effects downstream, depending on the speed of the FIFO.

stefanlinden:
Hi @jwmelto, thanks for the detailed code review, it is appreciated. As you can see my own background is mostly in c so a lot of the c++ nuances are new to me.
I've been doing some performance analyses between the 'new' version with buffering and the one 'old' implementation without buffering and indeed there is no significant difference based on GNSS-SDR's own timer.
Therefore, for simplicity's sake I will remove the buffer and apply the proposed changes that are still relevant.

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
stefanlinden: Put back in [77c682e](https://github.com/gnss-sdr/gnss-sdr/pull/584/commits/77c682e8b312c645d77e60289b0d67b33e67c633).

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.cc
stefanlinden: Obsolete after [77c682e](https://github.com/gnss-sdr/gnss-sdr/pull/584/commits/77c682e8b312c645d77e60289b0d67b33e67c633).

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.h
stefanlinden: Not applicable anymore after [77c682e](https://github.com/gnss-sdr/gnss-sdr/pull/584/commits/77c682e8b312c645d77e60289b0d67b33e67c633).

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.h
stefanlinden: Not applicable anymore after [77c682e](https://github.com/gnss-sdr/gnss-sdr/pull/584/commits/77c682e8b312c645d77e60289b0d67b33e67c633).

src/algorithms/signal_source/gnuradio_blocks/fifo_reader.h
stefanlinden: Thanks. For observers, also see: https://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file.

carlesfernandez:
Thanks for your work, the PR seems fine to me. Let's wait for @jwmelto  approval.

That's great because we want to make a new release in the following days, so all the changes in the current `next` branch get into the forthcoming GNU/Linux distro releases. I hope to get this PR in.

jwmelto (approved):
Looks great.  Thanks.

PR #587, opened 2022-02-22
correct name of SSE3 in Linux kernel for volk_profile SIMD detection

jmfriedt:
Correct name of SSE3 for using SIMD on volk-gnss. See https://github.com/google/cpu_features/pull/225

Signed-off-by: Jean-Michel Friedt <friedtj@free.fr>


carlesfernandez:
Thanks for the fix!

PR #593, opened 2022-04-03
Add missing #include <boost/exception/diagnostic_information.hpp>

AdrianBunk:
This fixes the build failures at
https://buildd.debian.org/status/logs.php?pkg=gnss-sdr&ver=0.0.16-1%2Bb2
```
/<<PKGBUILDDIR>>/src/algorithms/signal_source/adapters/limesdr_signal_source.cc: In constructor ‘LimesdrSignalSource::LimesdrSignalSource(const ConfigurationInterface*, const string&, unsigned int, unsigned int, Concurrent_Queue<std::shared_ptr<pmt::pmt_base> >*)’:
/<<PKGBUILDDIR>>/src/algorithms/signal_source/adapters/limesdr_signal_source.cc:100:67: error: ‘diagnostic_information’ is not a member of ‘boost’
  100 |                     LOG(WARNING) << "Boost exception: " << boost::diagnostic_information(e);
      |                                                                   ^~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/algorithms/signal_source/adapters/limesdr_signal_source.cc
+++ b/src/algorithms/signal_source/adapters/limesdr_signal_source.cc
@@ -19,6 +19,7 @@
 #include "gnss_frequencies.h"
 #include "gnss_sdr_string_literals.h"
 #include "gnss_sdr_valve.h"
+#include <boost/exception/diagnostic_information.hpp>
 #include <glog/logging.h>
 #include <gnuradio/blocks/file_sink.h>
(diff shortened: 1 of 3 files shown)


carlesfernandez:
This fix is already in the `next` branch, but I'll accept this PR since the `main` branch should be always usable with the latest GNU Radio version. But this also means that we need to make a new release soon :-)

Thanks for your contribution.

PR #599, opened 2022-04-26
Move duplicated files in asn-supl/ and asn-rrl/ to a new types/ subdir.

orlando017:
Signed-off-by: Will Silberman [wsilberm@google.com](mailto:wsilberm@google.com)

Move duplicated files in asn-supl/ and asn-rrl/ to a new types/ subdir. Note that in some cases, there are slight differences between the versions from the two folders, such as function argument names or in some cases data types; eg here's the diff between asn-supl/INTEGER.c and asn-rrlp/INTEGER.c:

54c54
<      * (Remove too int64_t sign extension, remove some first 0x00 bytes)
---
>      * (Remove too long sign extension, remove some first 0x00 bytes)
226c226
<     /* Output in the int64_t xx:yy:zz... format */
---
>     /* Output in the long xx:yy:zz... format */
421c421
<     int64_t sign = 1LL;
---
>     int64_t sign = 1;
1057c1057
<              * able to >sizeof(int64_t) bytes,
---
>              * able to >sizeof(long) bytes,
1085c1085
<                     /* Still cannot fit the int64_t */
---
>                     /* Still cannot fit the long */

I suspect such changes are inadvertent from after the code was duplicated, so could probably use some guidance on which versions to use.

Having the duplicated symbols is error prone, and causes issues when linking with some compilers.

I tested this by building and running gnss-sdr per https://gnss-sdr.org/my-first-fix/, and the program ran fine as far as I could tell.

### DIFF
--- a/src/core/libs/supl/types/BIT_STRING.c
+++ b/src/core/libs/supl/types/BIT_STRING.c
@@ -1,0 +1,207 @@
+/*-
+ * SPDX-FileCopyrightText: (c) 2003, 2004 Lev Walkin <vlm@lionet.info>. All rights reserved.
+ * SPDX-License-Identifier: BSD-1-Clause
+ */
+#include <BIT_STRING.h>
+#include <asn_internal.h>
+
+/*
+ * BIT STRING basic type description.
+ */
+static ber_tlv_tag_t asn_DEF_BIT_STRING_tags[] = {
+    (ASN_TAG_CLASS_UNIVERSAL | (3 << 2))};
+static asn_OCTET_STRING_specifics_t asn_DEF_BIT_STRING_specs = {
+    sizeof(BIT_STRING_t), offsetof(BIT_STRING_t, _asn_ctx), ASN_OSUBV_BIT};
+asn_TYPE_descriptor_t asn_DEF_BIT_STRING = {
+    "BIT STRING",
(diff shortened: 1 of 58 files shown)


orlando017:
The commit is missing the changes to the CMakeLists.txt and deleting the duped files from asn-rrlp/ and asn-supl/. I'm working on adding them to the pull request in a follow-up commit.

jwmelto:
This isn't really reviewable, since every file shows up as new. Also the "duplicate" files weren't removed, so the value of this PR is limited.

Your building didn't exercise your changes, since there is no CMakeFile.txt changes to use them. Maybe try `git mv` from one source to your new directory, and `git rm` from the other source. Once the old files are gone, your rest will be more useful.

To your question, the explicitly sized types (e.g., `int64_t`) are preferable in this situation.

jwmelto:
Also, I'd recommend making a distinct branch, rather than committing to your repo's `next` branch. It's cleaner that way. Normally, your repo `next` should either be an exact copy of upstream, or a few commits behind. It's not good practice to have your `next` diverge from upstream.

orlando017:
Thanks! I created another pull request https://github.com/gnss-sdr/gnss-sdr/pull/600 to replace this one. Closing this request.

PR #600, opened 2022-04-26
Move duplicated files in asn-supl/ and asn-rrl/ to a new types/ subdir

orlando017:
Signed-off-by: Will Silberman [wsilberm@google.com](mailto:wsilberm@google.com)

Move duplicated files in asn-supl/ and asn-rrl/ to a new types/ subdir. Note that in some cases, there are slight differences between the versions from the two folders, such as function argument names or in some cases data types; eg here's the diff between asn-supl/INTEGER.c and asn-rrlp/INTEGER.c:

```
54c54
<      * (Remove too int64_t sign extension, remove some first 0x00 bytes)
---
>      * (Remove too long sign extension, remove some first 0x00 bytes)
226c226
<     /* Output in the int64_t xx:yy:zz... format */
---
>     /* Output in the long xx:yy:zz... format */
421c421
<     int64_t sign = 1LL;
---
>     int64_t sign = 1;
1057c1057
<              * able to >sizeof(int64_t) bytes,
---
>              * able to >sizeof(long) bytes,
1085c1085
<                     /* Still cannot fit the int64_t */
---
>                     /* Still cannot fit the long */
```
I suspect such changes are inadvertent from after the code was duplicated. Per jwmelto's suggestion in https://github.com/gnss-sdr/gnss-sdr/pull/599, I'm prioritizing the file versions in asn-supl, which. Ibelieve are all explicitly sized types.

Having the duplicated symbols is error prone, and causes issues when linking with some compilers.

I tested this by building and running gnss-sdr per https://gnss-sdr.org/my-first-fix/, and the program ran fine as far as I could tell.


carlesfernandez:
Dear @orlando017,

It seems that `src/core/libs/supl/types/*` is missing.

I like the idea of this PR, though. That code was generated automatically long time ago and then retouched to make modern compilers and linters happier, but it still does its job and retrieves A-GPS data from the net.  Since it will not be regenerated again (at least for that SUPL version), I'm in favor of making those changes if they allow more portability. 

For the duplicated files with small differences, I would retain those with `int64_t ` instead of `long`. We overlooked that.

orlando017:
Thank you! Fixed, added back the files in types/. All tests except for clang-tidy are now succeeding; I'm not sure what's going wrong there, from the 'checks' tab it looks like the check is still running.

Is there a way to restart the clang tidy check (or all checks)?

src/core/libs/supl/CMakeLists.txt
```diff
@@ -55,9 +62,10 @@ target_include_directories(core_libs_supl
         ${CMAKE_CURRENT_SOURCE_DIR}
         ${CMAKE_SOURCE_DIR}/src/core/libs/supl/asn-supl
         ${CMAKE_SOURCE_DIR}/src/core/libs/supl/asn-rrlp
+        ${CMAKE_SOURCE_DIR}/src/core/libs/supl/types
```
jwmelto, line 65: This change is consistent with the surrounding lines, however there is an opportunity to improve all of them (I usually fix too much, so feel free to ignore this). A more maintainable change would be 
```suggestion
        ${CMAKE_CURRENT_SOURCE_DIR}/types
```

src/core/libs/supl/CMakeLists.txt
```diff
@@ -55,9 +62,10 @@ target_include_directories(core_libs_supl
...
 )
 
 set_target_properties(core_libs_supl PROPERTIES
     LINKER_LANGUAGE C
-    INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR};${CMAKE_SOURCE_DIR}/src/core/libs/supl/asn-supl;${CMAKE_SOURCE_DIR}/src/core/libs/supl/asn-rrlp;${GNUTLS_INCLUDE_DIR}"
+    INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR};${CMAKE_SOURCE_DIR}/src/core/libs/supl/asn-supl;${CMAKE_SOURCE_DIR}/src/core/libs/supl/asn-rrlp;${CMAKE_SOURCE_DIR}/src/core/libs/supl/types;${GNUTLS_INCLUDE_DIR}"
```
jwmelto, line 70: Similarly, consider
```suggestion
    INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR};${CMAKE_CURRENT_SOURCE_DIR}/asn-supl;${CMAKE_CURRENT_SOURCE_DIR}/asn-rrlp;${CMAKE_CURRENT_SOURCE_DIR}/types;${GNUTLS_INCLUDE_DIR}"
```

Now, I'd have to prove this out, but I'm pretty sure that this is equivalent (and more maintainable):
```suggestion
    INTERFACE_INCLUDE_DIRECTORIES 
        ${CMAKE_CURRENT_SOURCE_DIR}
        ${CMAKE_CURRENT_SOURCE_DIR}/asn-supl
        ${CMAKE_CURRENT_SOURCE_DIR}/asn-rrlp
        ${CMAKE_CURRENT_SOURCE_DIR}/types
        ${GNUTLS_INCLUDE_DIR}
```

src/core/libs/supl/CMakeLists.txt
```diff
@@ -12,15 +12,21 @@ file(GLOB ASN_SUPL_SOURCES "${CMAKE_SOURCE_DIR}/src/core/libs/supl/asn-supl/*.c"
 list(SORT ASN_SUPL_SOURCES)
 file(GLOB ASN_SUPL_HEADERS "${CMAKE_SOURCE_DIR}/src/core/libs/supl/asn-supl/*.h")
 list(SORT ASN_SUPL_HEADERS)
+file(GLOB ASN_TYPES_SOURCES "${CMAKE_SOURCE_DIR}/src/core/libs/supl/types/*.c")
```
jwmelto, line 15: As a general rule, CMake recommends _against_ using **glob** patterns. The one place where they are appropriate is with generated files. As @carlesfernandez kindly pointed out, this code was originally generated, which explains the pattern here.

If we are now moving to explicit maintenance, would it be better to explicitly enumerate the sources and their headers? I don't feel strongly about this, but hope that the maintainers will have an opinion.

jwmelto (approved):
Thanks for using proper **git** practices. It made this so much easier to review. Given that you are literally moving one set of files to another directory and deleting the duplicate files, I have no issues with this PR. For what that's worth.

src/core/libs/supl/CMakeLists.txt
carlesfernandez: That's nice, I will apply that before merging.

src/core/libs/supl/CMakeLists.txt
orlando017: Done.

src/core/libs/supl/CMakeLists.txt
orlando017: I'd rather leave as is for now if that's ok. It would be a very long list to write out.

carlesfernandez:
Thanks! Merging this PR into the `next` branch.

PR #602, opened 2022-05-01
Fix large glonass velocity error

vladisslav2011:
Subtract carrier offset from d_current_doppler when updating
current_syncro_data.

Signed-off-by: Vlad P <vladisslav2011@gmail.com>

### DIFF
--- a/src/algorithms/tracking/gnuradio_blocks/glonass_l1_ca_dll_pll_c_aid_tracking_cc.cc
+++ b/src/algorithms/tracking/gnuradio_blocks/glonass_l1_ca_dll_pll_c_aid_tracking_cc.cc
@@ -129,7 +129,7 @@ glonass_l1_ca_dll_pll_c_aid_tracking_cc::glonass_l1_ca_dll_pll_c_aid_tracking_cc
       d_code_phase_step_chips(0.0),
       d_carrier_doppler_hz(0.0),
       d_carrier_frequency_hz(0.0),
-      d_carrier_doppler_old_hz(0.0),
+      d_carrier_frequency_old_hz(0.0),
       d_carrier_phase_step_rad(0.0),
       d_acc_carrier_phase_cycles(0.0),
       d_code_phase_samples(0.0),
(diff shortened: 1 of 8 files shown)


jwmelto:
I see **what** you are doing, but need some explanation of **why** it is correct.

Seeing the same modification 12 times (6 times with 2 frequencies) begs for some more maintainable abstraction, but I don't fault you for not doing it.

Don't use `static_cast` unless you really need to. In this case, you are multiplying by a double, so there is no need to cast the other operand. Normal type promotion rules will do the right thing.

vladisslav2011:
Thanks for a suggestion.
I have looked at the code for one more time and I think it needs some refactoring instead of this quick fix. It will be better to use `d_carrier_frequency_hz` for carrier + doppler instead of `d_carrier_doppler_hz` and use `d_carrier_doppler_hz` for just doppler offset. This would eliminate some duplicate code. I'll do some tests and force-push a new version.
Another problem is that `msg_handler_preamble_index` never gets called (flowgraph issue?) and extended correlation never gets enabled. That's not a big problem as I have not seen frequent "Loss of lock..." messages like when tracking GPS L1 without extended correlation enabled.
I have tested `glonass_l2_ca_dll_pll_c_aid_tracking_cc` with this quick fix and can confirm that gnss-sdr-monitor shows correct doppler at L2. I can't confirm that velocity is correct as I have not obtained a fix at L2 due to bad antenna (simple dipole without filter/LNA).

`static_cast` was copied from here https://github.com/gnss-sdr/gnss-sdr/blob/fe2ef843348ff45d3b8a2152365ea215b5af5371/src/algorithms/tracking/gnuradio_blocks/glonass_l1_ca_dll_pll_c_aid_tracking_cc.cc#L215
Parenthesis around multiplication are not required too.

vladisslav2011:
Updated.
Short test with old baseband recording:
https://www.youtube.com/watch?v=j-QQT2aA3CM

Could you suggest comments to add to code?

jwmelto:
> Parenthesis around multiplication are not required too.

Not required, but I would encourage leaving that rather than relying on people to understand order of operator precedence.

carlesfernandez:
Thanks!

PR #604, opened 2022-05-17
Fix GLONASS extended correlator

vladisslav2011:
1. Emit a message from preamble detection branch from telemetry decoder
2. Connect flowgraph message ports to pass the message from telemetry
decoder block to a dll-pll-c-aid-tracking block
3. Name message ports correctly as the massage is passed in samples, not
seconds

Small demonstration: https://www.youtube.com/watch?v=eTc6ZX22jEo

### DIFF
--- a/src/algorithms/channel/adapters/channel.cc
+++ b/src/algorithms/channel/adapters/channel.cc
@@ -128,6 +128,10 @@ void Channel::connect(gr::top_block_sptr top_block)
 
     // Message ports
     top_block->msg_connect(nav_->get_left_block(), pmt::mp("telemetry_to_trk"), trk_->get_right_block(), pmt::mp("telemetry_to_trk"));
+    if ((trk_->get_right_block()->name() == "glonass_l1_ca_dll_pll_c_aid_tracking_cc") || (trk_->get_right_block()->name() == "glonass_l2_ca_dll_pll_c_aid_tracking_cc"))
+        {
+            top_block->msg_connect(nav_->get_left_block(), pmt::mp("preamble_timestamp_samples"), trk_->get_right_block(), pmt::mp("preamble_timestamp_samples"));
+        }
     DLOG(INFO) << "tracking -> telemetry_decoder";
 
     // Message ports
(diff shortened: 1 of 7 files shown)


carlesfernandez:
Thanks!

src/algorithms/channel/adapters/channel.cc
```diff
@@ -128,6 +128,10 @@ void Channel::connect(gr::top_block_sptr top_block)
 
     // Message ports
     top_block->msg_connect(nav_->get_left_block(), pmt::mp("telemetry_to_trk"), trk_->get_right_block(), pmt::mp("telemetry_to_trk"));
+    if ((trk_->get_right_block()->name() == "glonass_l1_ca_dll_pll_c_aid_tracking_cc") || (trk_->get_right_block()->name() == "glonass_l2_ca_dll_pll_c_aid_tracking_cc"))
```
jwmelto, line 131: I know this is already merged, but this is terribly brittle. The adapter should probably explicitly enable the processing of this message based on a config file entry, rather than tying this to one specific block name (which might change in the future, silently breaking this code).

Better coupling should be found.

vladisslav2011: Hello.
Other possible solutions are:
1. Check if the tracking block has `telemetry_to_trk` message in port and if the telemetry decoder block has `telemetry_to_trk` message out port and only then make a connection. This may silently fail too if the message pmt symbol gets changed.
2. Wrap `msg_connect` into try..catch. This would, of course, silently fail if anything gets changed.
3. Match `trk_->get_right_block()->name()` against /glonass_l._ca_dll_pll_c_aid_tracking_.c/ regexp. This would cover all 4 blocks and, of course, would have problems if block name/message pmt symbol gets changed.
I'll try to check the config file for presence of `Acquisition_[1,2]G.coherent_integration_time_ms` and make a connection in this case only.
And it looks, like I forgot to update `glonass_l1_ca_dll_pll_c_aid_tracking_sc` and `glonass_l2_ca_dll_pll_c_aid_tracking_sc` blocks.
Thanks for a review.

vladisslav2011: @jwmelto, I've opened a new PR https://github.com/gnss-sdr/gnss-sdr/pull/625, following your suggestions. Take a look, please.

PR #605, opened 2022-05-17
Osmosdr bandwidth selection

vladisslav2011:
Make it possible to override osmosdr default filter bandwidth.
Add new config file parameter:

SignalSource.if_bw=&lt;bandwidth in Hz&gt;

### DIFF
--- a/src/algorithms/signal_source/adapters/osmosdr_signal_source.cc
+++ b/src/algorithms/signal_source/adapters/osmosdr_signal_source.cc
@@ -42,6 +42,7 @@ OsmosdrSignalSource::OsmosdrSignalSource(const ConfigurationInterface* configura
       gain_(configuration->property(role + ".gain", 40.0)),
       if_gain_(configuration->property(role + ".if_gain", 40.0)),
       rf_gain_(configuration->property(role + ".rf_gain", 40.0)),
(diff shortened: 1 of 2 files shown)


src/algorithms/signal_source/adapters/osmosdr_signal_source.cc
```diff
@@ -93,6 +94,7 @@ OsmosdrSignalSource::OsmosdrSignalSource(const ConfigurationInterface* configura
                     osmosdr_source_->set_gain(gain_, 0);
                     osmosdr_source_->set_if_gain(rf_gain_, 0);
                     osmosdr_source_->set_bb_gain(if_gain_, 0);
+                    osmosdr_source_->set_bandwidth(if_bw_, 0);
```
jwmelto, line 97: What happens if you set the bandwidth to 0? It's unclear if this is requiring a bandwidth to be set, or truly "override(s) the default". I would imagine that for a GNSS application, a default bandwidth (see line 45) could be asserted.

vladisslav2011: Thanks.
That should not depend on `this->AGC_enabled_` state.
I think, this line is not required as `osmosdr_source` sets optimal bandwidth by default and we should not touch it if we don't have `Signal_source.if_bw` option in a config file.

Issue #621, opened 2022-06-17
no BDS satellites with PRN > 30 can be acquired and tracked


peiyzhou:
Hi! I was using gnss-sdr to decode my collected raw IF datasets with BDS B1I frequency. For the channels settings:

Channels_B1.count=40
Channels.in_acquisition=20
Channel.signal=B1

I was able to decode BDS B1I satellites for PRN smaller than 30 (from both BDS-2&3), but was not able to acqusite BDS satellites with PRN >30, which should be available there.  I was wondering if there are some settings related to PRNs which should be increased to processing satellites with larger PRNs?

Thank you!

peiyzhou:
it seems the PRNs are set to <32 in the function beidou_b1i_code_gen_int(), which is used for local PRN code generation.

vladisslav2011:
There are no specific settings.
You can see PRN 32 tracked at B1I here:
https://www.youtube.com/watch?v=BIf4bZthvUE
And PRN 33, PRN 40, PRN 56 at B3I here:
https://www.youtube.com/watch?v=jwFO56G27vk

minecraft2048:
> 

Off topic, but I wonder.. is the redacted coordinates an upstream feature?

peiyzhou:
> There are no specific settings. You can see PRN 32 tracked at B1I here: https://www.youtube.com/watch?v=BIf4bZthvUE And PRN 33, PRN 40, PRN 56 at B3I here: https://www.youtube.com/watch?v=jwFO56G27vk

B1I and B3I use different methods for generating PRN code. PRN code for B1I is only implemented for BDS PRN<=32 (function beidou_b1i_code_gen_int()), and PRN codes for all BDS B3I satellites are implemented.

peiyzhou:
> There are no specific settings. You can see PRN 32 tracked at B1I here: https://www.youtube.com/watch?v=BIf4bZthvUE And PRN 33, PRN 40, PRN 56 at B3I here: https://www.youtube.com/watch?v=jwFO56G27vk

Anyways, I use the current code to process B1I observations, including code and carrier phase, and use rtklib to obtain a BDS B1I only short baseline RTK solution. I was not able to fix the interger ambiguity, although the accuracy of the baseline solution is at cm-level. did you try to use the B1I carrier phase for RTK?? 
Thanks

vladisslav2011:
@minecraft2048
Just a joke. Nevermind.

vladisslav2011:
[...]

PR #624, opened 2022-06-26
Implement processing of BeiDou PRN34..PRN63 B1I signal

vladisslav2011:
Implemented processing of BeiDou PRN34..PRN63.
Added missing initial values and updated calculation according to http://www.beidou.gov.cn/xt/gfxz/201902/P020190227593621142475.pdf
Removed unused array.
Updated RTKLIB MAXPRNBDS to 63.
Tested with HackRF and GPS antenna. Works.
Close https://github.com/gnss-sdr/gnss-sdr/issues/621


vladisslav2011:
Help needed.
It looks, like RTKLIB does out-of-bounds access in position_test on MacOS. I do not have a Mac to debug this crash (under linux everything works well. Even memcheck does not find anything suspicious).
Anybody, having a Mac, run position_test under gdb and obtain a backtrace, please.

vladisslav2011:
I've fixed RTKLIB MacOS Ci failure (stack overflow). It should be ready for review once CI passes.

carlesfernandez:
Wow! This change should be extended to the whole RTKLIB, but this is something we can address down the road.  This PR  looks good! I confirm that it fixes the bug in MacOS.

PR #626, opened 2022-06-27
Fix clang-tidy CI failure

vladisslav2011:
Fix clang-tidy CI failure.

PR #627, opened 2022-06-27
Simplify configuration

vladisslav2011:
Make it possible to specify signal source per channel group.
Example:

;Set GPS L1 C/A channels RF channel ID to 1
Channels_1C.RF_channel_ID=1

instead of

;Set GPS L1 C/A channels RF channel ID to 1
Channel0.RF_channel_ID=1
Channel1.RF_channel_ID=1
Channel2.RF_channel_ID=1
....

### DIFF
--- a/src/core/receiver/gnss_flowgraph.cc
+++ b/src/core/receiver/gnss_flowgraph.cc
@@ -1010,7 +1010,8 @@ int GNSSFlowgraph::connect_signal_conditioners_to_channels()
 
             try
                 {
-                    selected_signal_conditioner_ID = configuration_->property("Channel" + std::to_string(i) + ".RF_channel_ID", 0);
+                    selected_signal_conditioner_ID = configuration_->property("Channels_" + channels_.at(i)->get_signal().get_signal_str() + ".RF_channel_ID", 0);
+                    selected_signal_conditioner_ID = configuration_->property("Channel" + std::to_string(i) + ".RF_channel_ID", selected_signal_conditioner_ID);
                 }
             catch (const std::exception& e)
                 {


src/algorithms/signal_source/libs/rtl_tcp_dongle_info.cc
```diff
@@ -21,8 +21,6 @@
 #include <string>
 #include <vector>
 
-using boost::asio::ip::tcp;
```
jwmelto, line 24: this change appears in all of your PRs, and is related to none of them. Are you starting from **next**?

vladisslav2011: Yes. I'm starting from next.
This change is separated into dedicated PR https://github.com/gnss-sdr/gnss-sdr/pull/626 now.

src/core/receiver/gnss_flowgraph.cc
```diff
@@ -1010,7 +1010,8 @@ int GNSSFlowgraph::connect_signal_conditioners_to_channels()
 
             try
                 {
-                    selected_signal_conditioner_ID = configuration_->property("Channel" + std::to_string(i) + ".RF_channel_ID", 0);
+                    selected_signal_conditioner_ID = configuration_->property("Channels_" + channels_.at(i)->get_signal().get_signal_str() + ".RF_channel_ID", 0);
+                    selected_signal_conditioner_ID = configuration_->property("Channel" + std::to_string(i) + ".RF_channel_ID", selected_signal_conditioner_ID);
```
jwmelto, line 1014: I don't know if there is any such documentation in the code, but the precedence implemented here should be documented somewhere, I think. Otherwise, this is a nice change.

vladisslav2011: I agree. We should document a new option and the fact, that an old option overrides the new one. I'll fork https://github.com/gnss-sdr/geniuss-place and open a PR.

PR #634, opened 2022-07-14
Limit BeiDou GEO satellites extend_correlation_symbols to 2

vladisslav2011:
...to improve tracking performance

Signed-off-by: Vlad P <vladisslav2011@gmail.com>

### DIFF
--- a/src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
+++ b/src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
@@ -109,6 +109,7 @@ dll_pll_veml_tracking::dll_pll_veml_tracking(const Dll_Pll_Conf &conf_)
       d_state(0),                 // initial state: standby
       d_current_prn_length_samples(static_cast<int32_t>(d_trk_parameters.vector_length)),
       d_extend_correlation_symbols_count(0),
+      d_extend_correlation_symbols(d_trk_parameters.extend_correlation_symbols),
       d_cn0_estimation_counter(0),
       d_carrier_lock_fail_counter(0),
       d_code_lock_fail_counter(0),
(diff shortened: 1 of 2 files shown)


src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
```diff
@@ -507,6 +508,7 @@ dll_pll_veml_tracking::dll_pll_veml_tracking(const Dll_Pll_Conf &conf_)
 
     d_multicorrelator_cpu.init(static_cast<int>(2 * d_trk_parameters.vector_length), d_n_correlator_taps);
 
+    d_extend_correlation_symbols = d_trk_parameters.extend_correlation_symbols;
```
carlesfernandez, line 511: This could be in the initializer list

src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
```diff
@@ -652,6 +654,7 @@ void dll_pll_veml_tracking::start_tracking()
     Signal_[0] = d_acquisition_gnss_synchro->Signal[0];
     Signal_[1] = d_acquisition_gnss_synchro->Signal[1];
     Signal_[2] = d_acquisition_gnss_synchro->Signal[2];
+    d_extend_correlation_symbols = d_trk_parameters.extend_correlation_symbols;
```
carlesfernandez, line 657: Duplicated?

src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
vladisslav2011: Thanks. I agree. The initialization may be skipped at all as `d_extend_correlation_symbols` is expected to be always initialized before use in `dll_pll_veml_tracking::start_tracking()`. But static analyzer may still emit warnings...
Wait a minute. I'll move it to initializer list and update a PR.

src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
vladisslav2011: No. That's reinitialization to default value in case, this channel was tracking GEO satellite, lost lock and got switched to MEO/IGSO satellite. It can go to `else` branch https://github.com/gnss-sdr/gnss-sdr/blob/1043b02e28efdc510b8308ed82c1cbfd5599ebc9/src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc#L775-L778, but I decided to not write duplicate code.

carlesfernandez:
Thanks!

PR #641, opened 2022-07-29
improve string handling to avoid buffer overrun

jwmelto:
I was required to perform a static code analysis of GNSS-SDR. There was one critical error identified.
image001

The fundamental issue is that if `strncpy` truncates the string, the result is not null-terminated, potentially leading to a buffer overrun. However, the specific error is that a substring is copied using the size of the whole string as a boundary. Underlying is the inherent lack of safety in using null-terminated strings. C++ addressed this issue (slowly) by creating a robust, efficient, and safe string class. Still, C-strings abound, and lots of legacy C code was just compiled with a C++ compiler without improvement.

I naively thought that I could easily refactor the offending portion of code to not use strncpy and friends, but it turns out that I found myself in one of those collections of "legacy C code compiled with a C++ compiler". 

Believe it or not, I tried very hard to limit the scope of this change; there was a lot more I could have done in this area. A major concern is that I don't know how to test this portion of the code. I would like to write unit tests, but we don't have the plumbing for low-level unit testing like this. I'm open to suggestions.

One point of orientation: the `file_t` structure and friends is (was) allocated using `malloc`; this means that no C++ constructors are called, making replacing fixed length character arrays with `std::string` was a lurking nightmare. So I carefully(? I think) found the allocation/deallocation code for _just that type_ and replaced with new/delete (actually using `std::unique_ptr`). This is **required** for classes with constructors to be used in the struct.

A lot of the code I ended up modifying should have been class member functions on the affected structure. That would be a useful improvement, but I didn't want to cloud this issue.

### DIFF
--- a/src/algorithms/libs/rtklib/rtklib.h
+++ b/src/algorithms/libs/rtklib/rtklib.h
@@ -1108,7 +1108,7 @@ typedef struct
 } serial_t;
 
 
-typedef struct
+struct file_t
 {                               /* file control type */
     FILE *fp = nullptr;         /* file pointer */
     FILE *fp_tag = nullptr;     /* file pointer of tag file */
@@ -1129,7 +1129,7 @@ typedef struct
     double speed = 0;           /* replay speed (time factor) */
     double swapintv = 0;        /* swap interval (hr) (0: no swap) */
     lock_t lock;                /* lock flag */
-} file_t;
+};
 
 
 typedef struct
--- a/src/algorithms/libs/rtklib/rtklib.h
+++ b/src/algorithms/libs/rtklib/rtklib.h
@@ -1109,26 +1109,26 @@ typedef struct
 
 
 typedef struct
-{                              /* file control type */
-    FILE *fp;                  /* file pointer */
-    FILE *fp_tag;              /* file pointer of tag file */
-    FILE *fp_tmp;              /* temporary file pointer for swap */
-    FILE *fp_tag_tmp;          /* temporary file pointer of tag file for swap */
-    char path[MAXSTRPATH];     /* file path */
-    char openpath[MAXSTRPATH]; /* open file path */
-    int mode;                  /* file mode */
-    int timetag;               /* time tag flag (0:off,1:on) */
-    int repmode;               /* replay mode (0:master,1:slave) */
-    int offset;                /* time offset (ms) for slave */
-    gtime_t time;              /* start time */
-    gtime_t wtime;             /* write time */
-    unsigned int tick;         /* start tick */
-    unsigned int tick_f;       /* start tick in file */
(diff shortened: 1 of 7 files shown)


jwmelto:
How important are the CI checks? In particular, **clang-tidy** wants me to change [rtklib_rtkcmn.cc:3105](https://github.com/gnss-sdr/gnss-sdr/blob/963b1cc63c192fedaeff9c5afefb3d6539efc59a/src/algorithms/libs/rtklib/rtklib_rtkcmn.cc#L3105) from
```c++
            auto sta = stas[np++];
```
to
```c++
            auto* sta = stas[np++];
```
See [SO](https://stackoverflow.com/a/12773328/9220132) for discussion. To me, the idea is to write "declare a temporary that is the type of `stas[np]` (which happens to be a pointer, or more properly `char[16]`). It seems counter-intuitive to me to say "declare a temporary pointer that is the type of `stas[np]` (and the SO sentiment appears to be on this side).

Similarly, **clang-tidy** wants to change (for example) [rtklib_stream.cc:1983](https://github.com/gnss-sdr/gnss-sdr/blob/963b1cc63c192fedaeff9c5afefb3d6539efc59a/src/algorithms/libs/rtklib/rtklib_stream.cc#L1983)
```c++
            if (ftp->proto) proto = "http"s;
```
to 
```c++
            if (ftp->proto) { proto = "http"s;
            }
```
but then **clang-format** would want to change it to
```c++
            if (ftp->proto) 
                { 
                    proto = "http"s;
                }
```
all of which diminishes from the readability of the code (doing it without the braces allowed **clang-format** to keep it on one line.

I can naively make the changes that **clang-tidy** wants, at the code of decreased readability, it strict compliance is a program goal.

carlesfernandez:
Dear @jwmelto,

many thanks for this pull request. While clang-tidy use to be useful and helps to keep everything "tidy", sometimes it gets quite picky, as it is in this case.

For those cases in which readability is compromised, you can use a `\\NOLINT` comment (see the [documentation](https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics)), which will suppress the warning for you.

jwmelto:
All checks (finally) passing

carlesfernandez:
Thanks for this PR!

I had to add a couple of things to make it work for C++11 compilers, using `Boost::filesystem` if `std::filesystem` is not available, and including `gnss_sdr_make_unique.h` to be able to use `std::make_unique`. Now it seems to work fine everywhere, so merging :-)

jwmelto:
sorry ... I've been building gnu-radio 3.10 and they are C++-17 minimum now, so I got sloppy. When do you think we might require new builds to use more modern tools? GNU Radio requires gcc >= 9

PR #642, opened 2022-07-30
Remove marketing blurb for snapcraft instructions

ShadowJonathan:
While I have some personal opinions about the base reasoning of snaps, the blurb that tells they're "universal" is categorically not true, as snaps require a lot of invasive changes that distros other than ubuntu (or very ubuntu-like) simply do not execute.

I think describing the reality - that snaps are best used on an ubuntu distro - will go further in addressing the needs of someone looking at this software, which is "how and where can i use this?", instead of becoming frustrated that snaps arent widely accepted.

`Signed-off-by: Jonathan de Jong <jonathan@automatia.nl>`


carlesfernandez:
Fair enough.

PR #644, opened 2022-07-30
Fixup MacOS install `cask` command

ShadowJonathan:
Encountered this while trying to install on macos, as newer versions of brew have deprecated the old way.

`Signed-off-by: Jonathan de Jong <jonathan@automatia.nl>`


carlesfernandez:
Thanks!

PR #646, opened 2022-08-01
improved precision of math constants

koloboxer:
The value of `TWO_N15` has an error of the order of $10^{-10}$. Current machines have the precision of double values around $10^{-16}$, so there is no sense to have such an error in a constant.

A change in the expression for `TWO_N16` may be not that important, but I think there is no sense to write a `double` variable using `float` variable syntax, which may also lead to a loss of precision.

Signed-off-by: Alexander V. Joura <ellfire@gmail.com>


carlesfernandez:
Thanks!

PR #647, opened 2022-08-03
Secure User Location Protocol fixes

koloboxer:
1. Member function definitions in `core/libs/gnss_sdr_supl_client.cc` did not have class name qualifiers, so I have added them.
2. Declarations of these functions in `core/libs/gnss_sdr_supl_client.h` did not have ampersands, also added those.
3. Function `Gnss_Sdr_Supl_Client::load_gnav_ephemeris_xml` used wrong ephemeris map (GPS instead of GLONASS)


carlesfernandez:
Thanks!

PR #651, opened 2022-08-11
add ZMQ signal source

jwmelto:
This change adds a new signal source type, **ZMQ_Signal_Source**. The motivation for this was to allow for existing GNU Radio flowgraphs to handle the data source control, potentially including various transformations, then publish samples via ZeroMQ. This **ZMQ_Signal_Source** could subscribe to that stream.

I also added a reusable `.item_type` to size converter, but did not retrofit other signal sources to use it, to minimize the scope of this change.

I originally implemented it with multichannel_file_source in mind, where multiple streams could be brought together; I had in mind synchronizing those sources as n-ary outputs. But I realized that GNSS-SDR has little need for that construct, so I simplified it to a single channel.


carlesfernandez:
Thanks! Documented [here](https://gnss-sdr.org/docs/sp-blocks/signal-source/#implementation-zmq_signal_source).

jwmelto:
oooh. I forgot about documentation! Thanks for cleaning up after me.

A couple of notes:
* there is an error in the GNU Radio documentation. The LSB of the `timeout` is in milliseconds, not microseconds. There was a breaking API change in ZMQ 3.0 that changed the units from us to ms. GNU Radio handles this conversion silently, but the API units are milliseconds.
* ZMQ requires a protocol for an endpoint (see `SignalSource.endpoint`). Also, for TCP endpoints, "localhost" doesn't do what you want or expect. You need to provide IP addresses, as in `tcp://127.0.0.1:55555`. Or, you might use an IPC, as in `ipc://@virtual-radio`

carlesfernandez:
Fixed! Thanks

PR #652, opened 2022-08-16
Wrong decoding of GLONASS navigation message parameter P1

koloboxer:
Current version of GNSS SDR produces values of P1 according to the following table

P1 | d_P1
00 | 15
01 | 30
10 | 45
11 | 60

However Glonass manual says that the table should be as follows

glonass-p1

(sorry, cannot find an english version of the manual).

Therefore I've changed conversion rules and tests, corresponding to these conversion rules


carlesfernandez:
Thanks!

Issue #653, opened 2022-08-16
CMake issues building with gnuradio 3.10


jwmelto:
GNU Radio moved from the unsupported log4cpp to spdlog in 3.10. The addressing of this in GNSS-SDR was, I fear, over-engineered. The current configuration will not compile on my CentOS 7 configuration. CentOS suffers from missing or old packages, so many things have to be built from source.

In the **FindSPDLOG.cmake** module, there is a blind assertion of a compile definition `-DSPDLOG_FMT_EXTERNAL=1`. This is incorrect. However, it's a symptom of a deeper issue.

When **spdlog** is installed, it provides CMake support files, such that `find_package(spdlog CONFIG)` should set all necessary variables. That we do `find_package(SPDLOG REQUIRED)` creates a need to write a redundant custom FindXXX.cmake module.

More to the point, as a consumer of GNU Radio, we shouldn't be specifying GNU Radio dependencies in the first place. It is the obligation of GNU Radio to expose their library dependencies through _their_ CMake modules.

I see a large ripple through the GNSS-SDR code-base where CMakeLists.txt have been expanded to add `spdlog::spdlog fmt::fmt` as link dependencies in all of our code that uses GNU Radio. I don't think this is needed. I'll be experimenting and see if I can reduce the tentacles caused by this approach.

ryanvolz:
As someone who has packaged GNU Radio and already had to figure out what's happening with `spdlog` and `libfmt`, hopefully I can offer some guidance. There was a period of time early on when GNU Radio 3.10 was not depending on `spdlog` correctly, and I think working around that bug is why `gnss-sdr` is having problems now.

If everything is working as intended (e.g with GR 3.10.3 at least), `gnss-sdr` should not need to add `-DSPDLOG_FMT_EXTERNAL=1`, and doing so causes problems for users with an `spdlog` that was built with the built-in `libfmt`. Also, there should be no circumstances where it would be necessary to independently link with `libfmt` or require it via CMake. `spdlog` will either depend on the external version or not, and be linked accordingly.

What I'm unsure of is whether or not it is necessary for `gnss-sdr` to depend on and link against `spdlog` directly. I think it is possible for a program to be built against GNU Radio and not itself link with `spdlog`, but I think it entirely depends on which GR headers it consumes. Most likely it is necessary.

[...]

PR #654, opened 2022-08-17
don't detect spdlog explicitly; let gnuradio handle it

jwmelto:
This pull request addresses #653.

Removing all the explicit assertion of **spdlog** and **fmt** dependencies works just fine with GNU Radio 3.10. I don't know about backwards compatibility, but since these were added in 3.10 there should be no problem.


jwmelto:
The ubuntu build failed, not finding log4cpp. I can put that part back in, but I'm wondering if we need a regression against 3.10.

I'm going to think about the 3.8 issues.

carlesfernandez:
Hi,

This seems to fail in:

* ArchLinux, which currently ships GNU Radio 3.10.2 (see [this GitLab job](https://gitlab.com/gnss-sdr/gnss-sdr/-/jobs/2894832443)).
* Fedora 38 Rawhide, which currently ships GNU Radio 3.10.3 (see [this GitLab job](https://gitlab.com/gnss-sdr/gnss-sdr/-/jobs/2894832455)).
*  OpenSUSE Tumbleweed, which currently ships GNU Radio 3.10.3 (see [this GitLab job](https://gitlab.com/gnss-sdr/gnss-sdr/-/jobs/2894832460)).
* Debian sid, which currently ships GNU Radio 3.10.3(see [this GitLab job](https://gitlab.com/gnss-sdr/gnss-sdr/-/jobs/2894832451)).
* Ubuntu 22.04, which ships 3.10.1 (see [this GitLab job](https://gitlab.com/gnss-sdr/gnss-sdr/-/jobs/2894832473)).

We need to dig a little bit more.

jwmelto:
As I suspected, this thread pulls long. Your GitLab jobs are instructive. I also came across dusty support for Debian 8, CentOS 7, and Ubuntu 14.04.

At issue is the difficulty in maintaining the GNSS-SDR shadow-versions of the GNU Radio provided CMake config file packages. What is the matrix of configurations that you believe need to be supported on **next**? The more and varied the versions, the harder this will be to maintain.

Note that GNU Radio 3.10 **_requires_** C++17, so older systems are going to have to be rev-locked at some point (and soon). It also requires, for GCC, version 9 or better. 

My recommendation is that, if you have to build from source, then plan on building GNU Radio too (if your distro is ancient, like mine is).

jwmelto:
A few notes. This change has exposed various issues outside of our control that need to be worked around:
* Ubuntu 20.04 installs `gnuradio-dev` (3.8.1) which in turn pulls in `volk2` (2.2.1). That version of volk has a dependency on `orc-0.4` that is not a package dependency. I updated the **main.yml** to pull it in
* Ubuntu 22.04 installs `gnuradio-dev` (3.10.1); the `gnuradio::gnuradio-blocks` target has an interface link dependency on `sndfile::sndfile`. This can't be found unless `libsndfile1-dev` is installed. That affects your GitLab job.
* CentOS 7 GitLab job needs to be updated to install and run `cmake3` instead of `cmake`. I removed support for CMake < 3.7

I suspect the other variants will be similar; I'll update here as I work through them.

jwmelto:
The MacOS failure is going to be tougher. I'm less sure how I'm going to reproduce that. Thus far, I've run in Docker containers, so the build is repeatable and clean every time.

jwmelto:
> The MacOS failure is going to be tougher. I'm less sure how I'm going to reproduce that. Thus far, I've run in Docker containers, so the build is repeatable and clean every time.

I ran on my Mac (12.4) and there were no errors

jwmelto:
Closing, because `next` has moved on, but I'm not sure the issue is fixed yet

PR #655, opened 2022-08-26
Glonass : avoid passing corrupt ephemeris and TOW measurements to observables

vladisslav2011:
Implemented Glonass ephemeris validation (much the same as with BeiDou).

Demonstration: https://www.youtube.com/watch?v=QvpAruQmgy8
GPS+Galileo for comparison: https://www.youtube.com/watch?v=WA-wMrXQyWg

### DIFF
--- a/src/algorithms/telemetry_decoder/gnuradio_blocks/glonass_l1_ca_telemetry_decoder_gs.cc
+++ b/src/algorithms/telemetry_decoder/gnuradio_blocks/glonass_l1_ca_telemetry_decoder_gs.cc
@@ -33,8 +33,6 @@
 #include <iostream>         // for cout
 #include <memory>           // for shared_ptr, make_shared
 
-#define CRC_ERROR_LIMIT 6
-
 
 glonass_l1_ca_telemetry_decoder_gs_sptr
 glonass_l1_ca_make_telemetry_decoder_gs(const Gnss_Satellite &satellite, const Tlm_Conf &conf)
@@ -61,7 +59,10 @@ glonass_l1_ca_telemetry_decoder_gs::glonass_l1_ca_telemetry_decoder_gs(
                             d_dump_mat(conf.dump_mat),
                             d_remove_dat(conf.remove_dat),
                             d_enable_navdata_monitor(conf.enable_navdata_monitor),
-                            d_dump_crc_stats(conf.dump_crc_stats)
+                            d_dump_crc_stats(conf.dump_crc_stats),
+                            d_ecc_errors_resync(conf.ecc_errors_resync),
+                            d_validator_thr(conf.validator_thr),
+                            d_validator_accept_first(conf.validator_accept_first)
 {
     // prevent telemetry symbols accumulation in output buffers
     this->set_max_noutput_items(1);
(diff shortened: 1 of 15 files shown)


ValZapod:
>does not give much improvement

But there is at least some 1 bit corrections if you "debug" with printf()?

Did you check that changing any [only] one bit in the string allows to get the original string?

vladisslav2011:
Have you seen the updated test: https://github.com/gnss-sdr/gnss-sdr/blob/10877bded1afd0ea57376181ebcaf455608f3d85/src/tests/unit-tests/system-parameters/glonass_gnav_crc_test.cc#L38-L44 ?
I'm checking, that every flipped data bit is being corrected by Glonass_Gnav_Navigation_Message::CRC_test. If not so, the test would fail and the CI would not pass.

It does not give much improvement because single bit errors do not happen frequently (in my tests). The string is either completely broken with many flipped bits and is uncorrectable (ephemeris corruption in this is handled by the validator) or the string does not have any flipped bits.

ZaquL:
Nice! Maybe to be completly sure, one should check those data? 


https://github.com/swift-nav/libswiftnav/blob/bafac07254507f975e20c222191431c0589d82f9/tests/check_decode_glo.c#L192-L197

If it all checks out (or maybe you already cehcked that!) LGTM.

vladisslav2011:
@ZaquL
> Maybe to be completly sure, one should check those data?

No problem. Test data added.

ZaquL:
Very nice! So, it all checkes out then!

vladisslav2011:
I have all tests passing on my local machine. Let's wait for the CI to complete.

carlesfernandez:
I’m not sure if this is the way to go. Ephemeris data have their own validation mechanism, and I feel this external validation is somewhat heuristic. In this sense, I prefer the implementation at https://github.com/gnss-sdr/gnss-sdr/pull/667, which seems to stick to the ICD.

However, this is a good idea that worth exploring it further, maybe as an anti-spoofing mechanism. Thanks for your work.

ValZapod:
>implementation at https://github.com/gnss-sdr/gnss-sdr/pull/667, which seems to stick to the ICD.

But it does NOT stick to ICD. Besides there are additional recommendations outside ICD from creaters of Glonass. Like I know about one: you are supposed to compare every two frames and compare the preparsed ephemerids from each (they are garanteed to be the same, except for tRecv from TK).

vladisslav2011:
@ValZapod 
It looks like this PR follows additional recommendations from creators of Glonass.

@carlesfernandez 
My tests are showing, that existing validation mechanism fails too often with real world signals (a signal from a vehicle-mounted antenna when driving through a city for example). This validator makes GNSS-SDR produce a better track with less position jumps and much more valid points.
I don't think, that this validator may help in case of spoofing. The GNSS-SDR will still switch to spoofed ephemeris after n frames.
Should I rebase the ECC error correction on top of next and open a separate PR?

carlesfernandez:
Yes, please, make another PR only with the ECC error correction.

ValZapod:
Is there any way to provide more internal documents about this? Or is it all TS//NF (perfectly secret that is)? Maybe it is in new CDMA docs of GLONASS somewhere?

vladisslav2011:
This is outdated and incomplete.
Maybe, I'll open a new PR, covering all telemetry decoders with tests and documentation updates.
Closing.

PR #657, opened 2022-08-26
add options for changing usrp device arguments

IntoPa:
Signed-off-by: Into Pääkkönen <into.paakkonen@aalto.fi>

These changes allow the user to change "num_recv_frames" and "recv_frame_size" transport parameters when using USRP devices. Default values for the parameters sometimes cause issues such as overflows when transporting data from USRP device to host. Detailed explanation of these parameters are given in USRP Manual Transport notes: https://files.ettus.com/manual/page_transport.html


carlesfernandez:
Thanks!

Merging. Also documented [here](https://gnss-sdr.org/docs/sp-blocks/signal-source/#implementation-uhd_signal_source).

PR #658, opened 2022-08-26
Position averaging

vladisslav2011:
While digging through source code, I noticed that `class Pvt_Conf` contains `uint32_t averaging_depth = 1;` field and there is unfinished averaging implementation in algorithms/PVT/libs/pvt_solution.cc, so I decided to finish the implementation.

New configuration parameter:
PVT.averaging_depth = 1 - Size of position averaging window. Set to a value <=1 to disable averaging.

### DIFF
--- a/src/algorithms/PVT/adapters/rtklib_pvt.cc
+++ b/src/algorithms/PVT/adapters/rtklib_pvt.cc
@@ -74,6 +74,10 @@ Rtklib_Pvt::Rtklib_Pvt(const ConfigurationInterface* configuration,
     // display rate
     pvt_output_parameters.display_rate_ms = bc::lcm(pvt_output_parameters.output_rate_ms, configuration->property(role + ".display_rate_ms", 500));
 
+    // averaging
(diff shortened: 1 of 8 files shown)


vladisslav2011:
Questions:
1. I have seen `bool PVT.flag_averaging` configuration parameter in some old config files. Should I implement processing of this parameter too? (Seems useless, as setting PVT.averaging_depth to 1 effectively disables the averaging)
2. I have implemented another improvement (not included here): a setting, that makes it possible to increase PVT solution rate while keeping PVT output rate low. It may be useful in combination with averaging. Should I open a PR featuring this option (and implementation of it's processing)?

vladisslav2011:
Demonstration: https://www.youtube.com/watch?v=JkJI6KhzN4o

carlesfernandez:
Superseded by the new [KF for PVT](https://gnss-sdr.org/docs/sp-blocks/pvt/#a-de-noising-kalman-filter-for-the-pvt-solution)

PR #662, opened 2022-09-06
Reduce gnss_syncro_monitor CPU load

vladisslav2011:
For some unknown reason gnss_synchro_monitor stays in a tight loop trying
to catch all syncros and sometimes consumes the whole CPU core.
Triggering the gnss_syncro_monitor from sample counter as it is done
with observables greatly reduces CPU usage from ~60...90% to <1%.

CPU usage.

Before:
gnss-syncro-monitor-bug

After:
gnss-synchro-monitor-fixed


src/core/monitor/gnss_synchro_monitor.cc
```diff
@@ -55,11 +55,12 @@ gnss_synchro_monitor::gnss_synchro_monitor(int n_channels,
 
 void gnss_synchro_monitor::forecast(int noutput_items __attribute__((unused)), gr_vector_int& ninput_items_required)
 {
-    for (int32_t channel_index = 0; channel_index < d_nchannels; channel_index++)
+    for (int32_t channel_index = 0; channel_index < d_nchannels - 1; channel_index++)
```
jwmelto, line 58: a comment above the block would be helpful; your change is not intuitive and a subsequent maintainer might not understand it.

I see the issue; by returning 0, it can cause this to spin when no samples are available (although the scheduler should still only be calling every 250ms). On the other hand, this says that a sample **must** be available in input channel `N-1` ... is that guaranteed? all other inputs would be starved until this was satisfied, if I understand the forecast method properly.

src/core/monitor/gnss_synchro_monitor.cc
```diff
@@ -91,6 +92,7 @@ int gnss_synchro_monitor::general_work(int noutput_items __attribute__((unused))
                         }
                 }
         }
+    consume(d_nchannels - 1, ninput_items[d_nchannels - 1]);
```
jwmelto, line 95: again, the logic is not obvious. Why is channel `N-1` not subject to decimation? Or any processing?

src/core/receiver/gnss_flowgraph.cc
```diff
@@ -206,7 +206,7 @@ void GNSSFlowgraph::init()
             udp_addr_vec.erase(std::unique(udp_addr_vec.begin(), udp_addr_vec.end()), udp_addr_vec.end());
 
             // Instantiate monitor object
-            GnssSynchroMonitor_ = gnss_synchro_make_monitor(channels_count_,
+            GnssSynchroMonitor_ = gnss_synchro_make_monitor(channels_count_ + 1,
```
jwmelto, line 209: Oh. This feels hack-y.

Inferring from your change, you believe the monitor block is spinning, calling `work()`; you modified `forecast()` to require 1 input on a "guard" input that you created here and attached to a counter source. If my inference of your assessment of the issue is correct, it is either a bug in GNU Radio or a (different?) logic error in the `forecast()` method.

jwmelto (commented):
I can believe that this improves performance (your evidence is compelling) but I'm not sure this is a quality fix. It may be; the case just hasn't been made.

src/core/monitor/gnss_synchro_monitor.cc
vladisslav2011: I've taken the logic from observables block, that does not hog the CPU when there are no synchros to process:
https://github.com/gnss-sdr/gnss-sdr/blob/next/src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc#L493-L501

src/core/monitor/gnss_synchro_monitor.cc
vladisslav2011: Same thing.
I'll add comment from observables block before this line.

src/core/monitor/gnss_synchro_monitor.cc
jwmelto: I guess it's a pattern. Yeah, add the comments and it is fine.

src/core/receiver/gnss_flowgraph.cc
vladisslav2011: Yes. I see, that monitor block is spinning calling work, but observables block, connected in parallel with monitor block does not. I've modified the monitor block, taking ideas from observables block source and it started consuming 2..3% of CPU core, as expected.
The same extra channel is added to observables block here https://github.com/gnss-sdr/gnss-sdr/blob/next/src/core/receiver/gnss_block_factory.cc#L331

jwmelto:
I still don't like the brittleness of this approach, but I am not sure there is a better one. I see you had to increment the channels to the ctor of the Acq monitor; the same change is needed (most likely) for the Tracking monitor.

Arguably, this could be moved to the ctor, so callers don't have to remember/know about this "extra" port (this is part of the brittleness; Or, you could just fix the tracking monitor as well.

vladisslav2011:
Thanks.
I've updated tracking monitor creation/connection.

I see at least 2 different solutions to this problem:
1. Pass synchros to the monitor as messages
2. Modify `gnss_synchro_monitor::general_work` to sleep for some time if all input buffers are empty

jwmelto:
> Thanks. I've updated tracking monitor creation/connection.
> 
> I see at least 2 different solutions to this problem:
> 
> 1. Pass synchros to the monitor as messages

This would be such a fundamental change to architecture, it's unlikely to be accepted.

> 2. Modify `gnss_synchro_monitor::general_work` to sleep for some time if all input buffers are empty

The default scheduler behavior should be to sleep for 250ms; however, I suspect that `forecast` is subverting this.

I don't have a problem with the time strobe being an input, but the _N+1_th is a maintenance burden. I don't know if there is a way to make a different "class" of input, to make it clear that it's different, or worst case, move it to the 0th input. It still creates a brittle reliance on external users to configure it correctly, and it's too easy to get wrong.

jwmelto:
@carlesfernandez can this be merged?

wkazubski:
> @carlesfernandez can this be merged?

I an using tkhis patch. It reduces CPU load significantly. Without the patch I had load burst to 100% when monitoring is enables, afer the patch is applied the CPU load is stable 25-30% while receiving GPS L1 signal with rtl-sdr dongle.

carlesfernandez:
We have implemented a more effective throttle control  in the Gnss_Synchro monitor. You can activate the decimation in the configuration file.

Thanks for your contribution!

PR #664, opened 2022-09-07
Shura current

koloboxer:
There were two class members in `gps_l1_ca_telemetry_decoder_gs` class

1. `d_bits_per_preamble`
2. `d_samples_per_preamble`

which were initialized from `GPS_CA_PREAMBLE_LENGTH_BITS` and never changed its value. It seemed not necessary, so I replaced them by just `GPS_CA_PREAMBLE_LENGTH_BITS` which is more readable: you can check the value of this `const` in IDE directly.


vladisslav2011:
Could you choose 'next' branch instead of 'main' as a target of your PR?

jwmelto:
in my opinion, this is not a desirable change. There is much duplicate code in the various decoders; I would think the preferred direction is towards greater abstraction, not more constellation/algorithm-unique duplication.

As far as I can tell, the motivation behind this is to make the IDE experience "nicer"? I believe that is misguided. Understanding the code should not depend upon knowing the precise value of the preamble values.

koloboxer:
@jwmelto , two reasons actually. First of all it does not make sense to have three variables for the same entity. `d_bits_per_preamble` and `d_samples_per_preamble` is the number of bits in the preamble, which is 8, and it is already defined by a constant `GPS_CA_PREAMBLE_LENGTH_BITS`. There is no sense to introduce two new class members for storing a value, which is not being changed in the process of calculation.

Just as an example... You have the constant `M_PI` in `math.h`. I hope that when writing your code, you do not introduce a new variable

    double my_pi = M_PI;

And if you do, I hope that at least you do not do it twice))

    double my_pi_1 = M_PI;
    double my_pi_2 = M_PI;

And the second reason is indeed the comfort in IDE, when you do not have to search the initialization of variables `d_bits_per_preamble` and `d_samples_per_preamble` just to find out that this is `GPS_CA_PREAMBLE_LENGTH_BITS`.

I do believe that these two class members exist only because of historical reasons.

PR #667, opened 2022-09-09
Hamming code correction method (Glonass)

koloboxer:
I have added Hamming code correction method for Glonass signal and google tests for this method. I have also replaced the call to `CRC_test` by the call to this new method, so that errors in messages are automatically corrected if possible.

The code works as described in  [Glonass ICD 5.1](https://russianspacesystems.ru/wp-content/uploads/2016/08/ICD_GLONASS_rus_v5.1.pdf), with only a single difference: the document says that if

    C_1 = C_2 = ... = C_7 = 0
    C_Sigma = 1

then the error is not recoverable. My experiments show that this case just means that all information bits are valid and there is a single error in `C_Sigma`. Therefore setting `C_Sigma` to `0` completely corrects the string.

Please, let me know if I am wrong, i.e. you can create a case

    C_1 = C_2 = ... = C_7 = 0
    C_Sigma = 1

by introducing 2 errors in a valid message.


vladisslav2011:
Have you seen my commit, that does the same thing https://github.com/gnss-sdr/gnss-sdr/pull/655/commits/533e107781ae423f3ad874d6fd10dce1cab512e8 and discussion here https://github.com/gnss-sdr/gnss-sdr/pull/655 ?

ValZapod:
> My experiments show that this case just means that all information bits are valid and there is a single error in `C_Sigma`

So the words "multiple errors" is a mistake? Anyway, mistake in checksum is not correctable, because that just means that there may be another mistake in some other bit of checksum and also mistakes in data and here you go, right checksum for wrong data. If you cannot get the checksum bits reliably, there is no point in trying to recover. That is a consensum for all error resilence AFAIK.

koloboxer:
> Have you seen my commit, that does the same thing [533e107](https://github.com/gnss-sdr/gnss-sdr/commit/533e107781ae423f3ad874d6fd10dce1cab512e8) and discussion here #655 ?

I have not seen it, thanks for the link. I will take a look and write back.

koloboxer:
> > My experiments show that this case just means that all information bits are valid and there is a single error in `C_Sigma`
> 
> So the words "multiple errors" is a mistake? Anyway, mistake in checksum is not correctable, because that just means that there may be another mistake in some other bit of checksum and also mistakes in data and here you go, right checksum for wrong data. If you cannot get the checksum bits reliably, there is no point in trying to recover. That is a consensus for all error resilence AFAIK.
> 
> Did you test with real satellites or just math?

It is just math, which is reflected in google tests. I take a valid string and generate all possible errors in this string. First all possible single bit errors, then all possible two bit errors.

Can you tell me, please, any reference to this consensus? Because I am not sure I understand the reasoning... If I get a message, which has all sums valid, it does not mean anything, because these correction codes are just probability based and I can change 3 bits in a valid message, keeping the check sums valid. So I guess there should be some research which says "OK, the strategy without recovering errors gives correct results in X % of cases which is larger than Y % for a strategy with error correction", or vice versa. Right?

ValZapod:
Evereywhere when parity bit does not pass you just remove the string. Why? Because even right out noise can pass this test then! And you even did not check it on real satellities.

vladisslav2011:
In real world, if you get a message, that requires error correction, chances, that next message will be uncorrectable, are very high. And chances, that you'll have correct navigation message decoded when you receive all required strings are very low.
I've made an IQ recording from a vehicle-mounted antenna (~26Gb of data) and did some tests. According to test results, enabling ephemeris validation and rejection gives noticeable performance improvement, enabling ECC does not give noticeable performance improvement, compared to 'main' branch.

ValZapod:
> enabling ECC does not give noticeable performance improvement, compared to 'main' branch.

You did not test in bad environment, did you? When ionosphere is chaotic. Or when there is some kind of passive error generator.

vladisslav2011:
Almost full recording was made in a "bad environment". The skyview was blocked by trees and buildings most of time. Only ~5 minutes of clean skyview (not continous). And, while driving through a city, there were some places with strong interference.
https://www.youtube.com/watch?v=QvpAruQmgy8

ValZapod:
> skyview was blocked by trees and buildings most of time.

Multipath really depends on whether you have a smart phase array like in a phone (bad, requires AI to really solve) or unidirectional antenna (your better case).

koloboxer:
> Everywhere when parity bit does not pass you just remove the string.

As far as I understand the parity bit is needed for correction only. Nothing more. If there is no intent to correct the message, the parity is not needed. If I have invalid parity, there are two options: single bit error error in parity or two or more bit error somewhere else. If the probability of single error is higher, I win by correcting the parity. Right?

koloboxer:
> In real world, if you get a message, that requires error correction, chances, that next message will be uncorrectable, are very high.
I am sorry, I did not understand the results. Was it better with correction in your experiments or without correction?

ValZapod:
How many times 1 bit was corrected?

vladisslav2011:
I had several runs with debug message enabled, but I have not implemented error counters. OK. I'll add counters (error-free strings, 1-bit errors and uncorrectable) and re-test.

vladisslav2011:
Some statistics:
```
Total GNSS-SDR run time: 243.291980 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:words_decoded=3547
Glonass_Gnav_Navigation_Message:words_corrected=121 (3.411%)
Glonass_Gnav_Navigation_Message:words_dropped=444 (12.518%)


Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 172.699069 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:words_decoded=3564
Glonass_Gnav_Navigation_Message:words_corrected=113 (3.171%)
Glonass_Gnav_Navigation_Message:words_dropped=464 (13.019%)


Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 172.801854 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:words_decoded=3541
Glonass_Gnav_Navigation_Message:words_corrected=115 (3.248%)
Glonass_Gnav_Navigation_Message:words_dropped=425 (12.002%)
```

I think, I should count successful decodes of navigation messages, containing words with 1-bit errors.

ValZapod:
> .. It is calculated against successful decodes instead of total words processed.

Yep, against words_decode. Still, 121 out of (3547+121+444) is a lot.

vladisslav2011:
'words` should be `strings` here.
Another 3 runs with more stats:
```
Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 170.387228 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3613
Glonass_Gnav_Navigation_Message:strings_corrected=105 (2.529%)
Glonass_Gnav_Navigation_Message:strings_dropped=434 (10.453%)
Glonass_Gnav_Navigation_Message:ephemeris_total=204
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=22 (10.784)%

Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 171.052081 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3563
Glonass_Gnav_Navigation_Message:strings_corrected=101 (2.462%)
Glonass_Gnav_Navigation_Message:strings_dropped=439 (10.699%)
Glonass_Gnav_Navigation_Message:ephemeris_total=203
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=20 (9.852)%

Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 172.540794 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3592
Glonass_Gnav_Navigation_Message:strings_corrected=104 (2.531%)
Glonass_Gnav_Navigation_Message:strings_dropped=413 (10.051%)
Glonass_Gnav_Navigation_Message:ephemeris_total=203
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=18 (8.867)%
```

> C_1 = C_2 = ... = C_7 = 0
> C_Sigma = 1

I have not tested for that specific case .

ValZapod:
>I have not tested for that specific case .

That is the question though. Arguably you may want to correct parity bit in this case, but it depends whether there is more noise that passes this vs real data. In first case we should follow ICD.

vladisslav2011:
I think, I have obtained more or less correct stats:
```
Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 171.495172 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3589
Glonass_Gnav_Navigation_Message:strings_corrected=102 (2.470%)
Glonass_Gnav_Navigation_Message:strings_dropped=439 (10.630%)
Glonass_Gnav_Navigation_Message:strings_dropped_csigma=1 (0.024%)
Glonass_Gnav_Navigation_Message:ephemeris_total=203
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=9 (4.433)%
```

vladisslav2011:
3 more runs:
```
Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 173.250314 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3533
Glonass_Gnav_Navigation_Message:strings_corrected=97 (2.415%)
Glonass_Gnav_Navigation_Message:strings_dropped=387 (9.634%)
Glonass_Gnav_Navigation_Message:strings_dropped_csigma=0 (0.000%)
Glonass_Gnav_Navigation_Message:ephemeris_total=200
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=9 (4.500)%

Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 174.797389 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3572
Glonass_Gnav_Navigation_Message:strings_corrected=113 (2.737%)
Glonass_Gnav_Navigation_Message:strings_dropped=444 (10.753%)
Glonass_Gnav_Navigation_Message:strings_dropped_csigma=1 (0.024%)
Glonass_Gnav_Navigation_Message:ephemeris_total=204
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=8 (3.922)%

Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 173.863523 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3520
Glonass_Gnav_Navigation_Message:strings_corrected=116 (2.869%)
Glonass_Gnav_Navigation_Message:strings_dropped=407 (10.067%)
Glonass_Gnav_Navigation_Message:strings_dropped_csigma=0 (0.000%)
Glonass_Gnav_Navigation_Message:ephemeris_total=201
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=10 (4.975)%

```

vladisslav2011:
More, than 50% of ephemeris with 1-bit errors corrected are passing validation. Not as bad, as I thought.
```
Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 171.794032 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3558
Glonass_Gnav_Navigation_Message:strings_corrected=104 (2.557%)
Glonass_Gnav_Navigation_Message:strings_dropped=405 (9.958%)
Glonass_Gnav_Navigation_Message:strings_dropped_csigma=0 (0.000%)
Glonass_Gnav_Navigation_Message:ephemeris_total=200
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=9 (4.500%)
Glonass_Gnav_Navigation_Message:ephemeris_dropped=11 (5.500%)
Glonass_Gnav_Navigation_Message:ephemeris_dropped_ecc=4 (2.000%)

Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 174.710891 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3580
Glonass_Gnav_Navigation_Message:strings_corrected=106 (2.583%)
Glonass_Gnav_Navigation_Message:strings_dropped=418 (10.185%)
Glonass_Gnav_Navigation_Message:strings_dropped_csigma=0 (0.000%)
Glonass_Gnav_Navigation_Message:ephemeris_total=202
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=10 (4.950%)
Glonass_Gnav_Navigation_Message:ephemeris_dropped=11 (5.446%)
Glonass_Gnav_Navigation_Message:ephemeris_dropped_ecc=5 (2.475%)

Current receiver time: 22 min 49 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 173.525739 [seconds]
GNSS-SDR program ended.
Glonass_Gnav_Navigation_Message:strings_decoded=3583
Glonass_Gnav_Navigation_Message:strings_corrected=111 (2.708%)
Glonass_Gnav_Navigation_Message:strings_dropped=405 (9.880%)
Glonass_Gnav_Navigation_Message:strings_dropped_csigma=0 (0.000%)
Glonass_Gnav_Navigation_Message:ephemeris_total=199
Glonass_Gnav_Navigation_Message:ephemeris_with_ecc=8 (4.020%)
Glonass_Gnav_Navigation_Message:ephemeris_dropped=11 (5.528%)
Glonass_Gnav_Navigation_Message:ephemeris_dropped_ecc=3 (1.508%)
```

ValZapod:
BTW, are those ephemeris better than what is available through Google SUPL? Higher precision? 

For RTK one needs ephemeris that are real, not predicted (as in SUPL or these), so that is why it is done after they are published... One of key reasons RTK is better.

vladisslav2011:
I have not tested Glonass in assisted mode, but I don't think, that published ephemeris are much different from broadcast ephemeris.
Getting ephemeris from Google SUPL may help to decrease time to first fix in case of GPS/Galileo?/BeiDou, but not Glonass. You'll always have to wait until the first ephemeris set will be received from minimum 4 satellites. On the other hand time to first fix is almost always less, than in case of unassisted GPG/Galileo/BeiDou.

For RTK it is always better to setup a base station in a place with known coordinates not far from the place, where actual measurements are to be done. You'll get corrected pseudoranges and phase measurements either in real-time or as a RINEX file to do post-processing.

ValZapod:
> For RTK it is always better to setup a base station

That is DGNSS (or even more cool WAAS, e.g., that uses a separate satellite and not yet supported in GNSS-SDR). I was talking about PPP, those are published, not predicted and you can only calculate a fix after those are published here: https://pppx.gdgps.net/ The ephermerids are rather different, since are not predicted and of course they have higher sigma, higher than 99.73%, I think.

>but not Glonass.

Are you sure you are not getting this cosmetic issue, where xml dump was not working? https://github.com/gnss-sdr/gnss-sdr/pull/647#issue-1327505145

vladisslav2011:
APPS is different thing. It is GPS-only. And, as I can see, they are offering post-processing only.
There is WAAS/SBAS telemetry decoder in GNSS-SDR, but there are no acquisition/tracking adapters (pcps_acquisition and dll_pll_veml_tracking blocks should work). So I have not test it too and can't confirm that it works with Glonass (at least ionospheric delay compensation should work).

> Are you sure you are not getting this cosmetic issue

No. Read the code of this method: https://github.com/gnss-sdr/gnss-sdr/blob/main/src/core/system_parameters/glonass_gnav_navigation_message.cc#L234 especially this line https://github.com/gnss-sdr/gnss-sdr/blob/main/src/core/system_parameters/glonass_gnav_navigation_message.cc#L363 and it's outer blocks.

ZaquL:
>strings_dropped_csigma=1 (0.024%)

So it is a very rare case. Does this string pass the validator if you correct the last bit (parity bit)?

I do not quite understand what does it mean "50% of ephemeris with 1-bit errors corrected are passing validation". Does it mean noise/garbage is passing validation?

>You'll always have to wait until the first ephemeris set will be received from minimum 4 satellites.

Can we implement some other AGPS service so that it would not be needed, or is it some kind of data that cannot be duplicated on the Internet? See, e.g. (only GPS, not GLONASS though), https://www.fruct.org/publications/fruct29/files/Vin.pdf

vladisslav2011:
> Does this string pass the validator if you correct the last bit (parity bit)?

This would be hard to test as this is a very rare case. One error per 4..5 runs.

> Does it mean noise/garbage is passing validation?

Noise/garbage may pass the validation only when `validator_accept_first` is enabled and only once per satellite. Chances, that two ephemeris sets with exactly the same set of flipped bits will be received without at least one ephemeris set with different set of flipped bits inbetween are very, very low.

> Can we implement some other AGPS service so that it would not be needed, or is it some kind of data that cannot be duplicated on the Internet?

No. It should receive time from each satellite before trying to solve for the PVT. Glonass satellites transmit time information once per 30 seconds (L1OF, L2OF signals). GPS satellites transmit time information once per 6 seconds.

ZaquL:
>Does it mean noise/garbage is passing validation?

I meant correction, sorry. If corrected ephemerids are not passing validation, that means the original data was noise, that just happens to have good checksum, which may be also noise?

> It should receive time from each satellite before trying to solve for the PVT

We can use GPS time and if we know speed and height even recalculate two relativity effects (GPS already has it corrected in transmitting frequncies, so maybe not even needed at least on GPS side). Also, one can use an atomic clock synced with Geneva right in the apartment. In the last case only 3 satelities are needed in fact to get 3D fix.

vladisslav2011:
> that means the original data was noise

The original data is not noise as it has correct preamble. This means, that there are more, than 2 bits flipped.

> In the last case only 3 satelities are needed in fact to get 3D fix.

Knowledge of absolute time does not help because delay from antenna to tracking/acquisition block is unknown and unstable.
3 satellites are enough to get 3D fix using single observation if approximate altitude is known and the constellation is good. 2 satellites are enough to get 3D fix using multiple observations if the receiver is fixed and a stable clock (ex. rubidium) is available.

ValZapod:
```
diff --git a/src/core/system_parameters/glonass_gnav_navigation_message.cc b/src/core/system_parameters/glonass_gnav_navigation_message.cc
index 5b02162a6..e1391d1d5 100644
--- a/src/core/system_parameters/glonass_gnav_navigation_message.cc
+++ b/src/core/system_parameters/glonass_gnav_navigation_message.cc
@@ -14[8](https://github.com/gnss-sdr/gnss-sdr/actions/runs/3036341256/jobs/4887523145#step:6:9),8 +148,[9](https://github.com/gnss-sdr/gnss-sdr/actions/runs/3036341256/jobs/4887523145#step:6:10) @@ bool Glonass_Gnav_Navigation_Message::hamming_correct(std::bitset<GLONASS_GNAV_S
     for (size_t m = 0; m < GLONASS_GNAV_HAMMING_CODE_BITS; ++m)
         {
             sum_bits = 0;
-            for (int i : GLONASS_GNAV_HAMMING_INDEX[m])
+            for (int i : GLONASS_GNAV_HAMMING_INDEX[m]) {
                 sum_bits += bits[i - 1];
+}
             C[m] = bits[m] ^ (sum_bits % 2);
         }
 
@@ -[18](https://github.com/gnss-sdr/gnss-sdr/actions/runs/3036341256/jobs/4887523145#step:6:19)1,8 +182,9 @@ bool Glonass_Gnav_Navigation_Message::hamming_correct(std::bitset<GLONASS_GNAV_S
             int highestNonZeroIdx = -1;
             for (int i = 0; i < GLONASS_GNAV_HAMMING_CODE_BITS - 1; ++i)  // C_Sigma should not be taken into account
                 {
-                    if (C[i] != 0)
+                    if (C[i] != 0) {
                         highestNonZeroIdx = i + 1;  // 1-based indices
+}
                 }
 
             if (shortSumC == 1)
```

ValZapod:
>The original data is not noise as it has correct preamble. This means that there are more, than 2 bits flipped.

Can you paste some hex editor style examples? This means there is some mistake (no info about cruptographic collisions) in ICD.

Okay, if that is that bad, we may as well add 

>My experiments show that this case just means that all information bits are valid and there is a single error in C_Sigma

If you confirm those cases above are like that (they have the C_Sigma flipped not almost noise with correct preamble).

ValZapod:
This can be closed got implemented in https://github.com/gnss-sdr/gnss-sdr/pull/694

PR #670, opened 2022-09-14
convert from vector to stream

jwmelto:
The ZMQ signal source most commonly will receive a vector of samples from a ZMQ publisher. This vector needs to be converted to a stream for GNSS-SDR use.


carlesfernandez:
Thanks! Merging

PR #671, opened 2022-09-20
Fix register unpacking for Labsat3W files

jeffmelville:
This change fixes a bug in the unpacking of Labsat 3 Wideband files when using the Labsat_Signal_Source. The original endian conversion loop includes a cast from char->uint64_t that (surprisingly) incurs a sign extension when the MSB of the char is set. ORing the unmasked uint64_t into the register can set undesired bits.

The changes replace the old endian conversion loop with a ``boost`` utility function.

Testing: Ran Labsat "UK Static" 3W demo through to PVT on a x86_64 (little endian) PC. I didn't have a big endian machine available for testing.

### DIFF
--- a/src/algorithms/signal_source/gnuradio_blocks/labsat23_source.cc
+++ b/src/algorithms/signal_source/gnuradio_blocks/labsat23_source.cc
@@ -21,6 +21,7 @@
 #include "INIReader.h"
 #include "command_event.h"
 #include "gnss_sdr_make_unique.h"
+#include <boost/endian/conversion.hpp>
 #include <gnuradio/io_signature.h>
 #include <algorithm>
 #include <array>
@@ -737,7 +738,8 @@ void labsat23_source::decode_ls3w_register(uint64_t input, std::vector<gr_comple
 {
     std::bitset<64> bs(input);
 
-    // Reverse, since register are written to file as 64-bit little endian words
+    // Earlier samples are written in the MSBs of the register. Bit-reverse the register
+    // for easier indexing. Note this bit-reverses individual samples as well for quant > 1 bit
     for (std::size_t i = 0; i < 32; ++i)
         {
             bool t = bs[i];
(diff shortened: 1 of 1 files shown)


jeffmelville:
Is "Signed-off-by" no longer sufficient? If not I need to set up gpg for my account

vladisslav2011:
Hello.

I think, it may be enough to replace
```
std::array<char, 8> memory_block{};
```
with
```
std::array<uint8_t, 8> memory_block{};
```
instead of adding more boost headers...

jeffmelville:
FWIW on my machine (Ubuntu 18.04), compiled for Release, these changes reduce the size of the `gnss-sdr` executable by 112 bytes (from 12277400 to 12277288)

carlesfernandez:
Hi,

`boost:endian` is a header-only library, so I expect it will be included with the `boost:headers` CMake target, already present in the `PUBLIC` links of the `signal_source_adapters` library.  So no need to modify the CMakeLists.txt file in that regard.

However, `boost::endian` seems to be part of Boost [since release 1.58.0](https://github.com/boostorg/endian). Our current minimum Boost version is 1.53.0, which is there for Centos 7. Ubuntu 14.04 shipped Boost 1.53.0, which still does not make it. Same for Debian 8, which shipped Boost 1.55.0.

A "hard" minimum version for Boost is 1.64.0, which is required by some [embedded systems](https://gnss-sdr.org/introducing-geniux-v2206/#geniux-rocko-22060). 

I guess just not building that source if the Boost version is too ancient would be an easy way to go. Is anyone using the `Labsat_Signal_Source` with those ancient distros? If so, please let us know.

A more elaborated fix would be to keep the removed code if the Boost version is too old, or even fixing it accounting for that MSB issue. But if the available Boost version is 1.58.0 or above, I'm in favour of using that utility function.

jeffmelville:
@carlesfernandez thanks for the feedback. Nice catch. I'll take a look and push an update.

PR #675, opened 2022-10-05
Fix the monitor console spam in case of unreachable destination address

### LINKED ISSUE #674: Adding monitor broadcast
The existing monitor pattern has some unfortunate implications.

1. Only a single port for any given IP address can receive the UDP packets. This means that exploitation of the data source is limited to one application (per computer) at a time.
2. **All** clients of the data must be known when the application is started. Adding a new client requires restarting the application
3. If a client is configured but not running, the output is filled with "unable to write" errors

A better pattern would be PUB/SUB or multicast. Because of the limitations of multicast, I would lean towards the PUB/SUB using ZeroMQ as an implementation. This allows both the client(s) and the server to be independent of each other in start-up order (or if at all).

Another complication in sharing this data is the communication of the ProtoBuf specifications to external clients. I don't have a solution for this; the `.proto` files are in the **docs/protobuf** directory, but this is not installed.

I think this pattern can co-exist with the current solution, but I'm looking for concurrence before implementing it.

vladisslav2011:
Use socket.send_to instead of socket.send

### DIFF
--- a/src/algorithms/PVT/libs/monitor_ephemeris_udp_sink.cc
+++ b/src/algorithms/PVT/libs/monitor_ephemeris_udp_sink.cc
@@ -58,11 +58,10 @@ bool Monitor_Ephemeris_Udp_Sink::write_galileo_ephemeris(const std::shared_ptr<G
     for (const auto& endpoint : endpoints)
         {
             socket.open(endpoint.protocol(), error);
-            socket.connect(endpoint, error);
 
             try
                 {
-                    if (socket.send(boost::asio::buffer(outbound_data)) == 0)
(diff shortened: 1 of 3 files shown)


jwmelto:
Looks good to me

jwmelto (approved):

jwmelto:
I don't think this is a fix for #674 but it a good improvement for the UDP sink

vladisslav2011:
Thanks.
I've removed "may fix" from PR description to prevent github from closing the ticket automatically.

carlesfernandez:
Thanks! Merging.

PR #680, opened 2022-10-28
acquisition: populate gnss_synchro.fs with acquisition sample rate

vladisslav2011:
This makes it possible to use GNSS-SDR-Monitor to monitor acquisition
process too.
Set AcquisitionMonitor.udp_port to the same value as Monitor.udp_port to
see acquisition results.
The acquisitions will be shown with a red telemetry mark and no tracking
data.
gnss-sdr-monitor


jwmelto:
I'm confused what this change does, besides fill in an uninitialized variable (which is a good thing).

vladisslav2011:
The GNSS-SDR-Monitor checks for initialization of `fs` member to confirm data validity https://github.com/vladisslav2011/gnss-sdr-monitor/blob/master/src/channel_table_model.cpp#L289
Initializing `fs` makes this check pass and the GNSS-SDR-Monitor starts showing both acquisition results and early tracking (before ephemeris reception and decoding).

jwmelto:
Looks good to me

PR #690, opened 2022-11-23
Fix input item type in config file and remove not working options

wkazubski:
:Fix input item type in gnss-sdr_GPS_L1_ishort.conf config file from cshort to gr_complex.
Remove SignalSource.enable_throttle_control=false option from hardware type config files as it does not work. In such case hardware determines sample rate and this option cannot be set to false.


carlesfernandez:
Thanks for your PR. Could you please [sign your commits](https://gnss-sdr.org/docs/tutorials/using-git/#sign-your-commits)?

PR #699, opened 2023-01-06
Flush NMEA output

jwmelto:
The NMEA file appears to be the only place where satellite azimuth and elevation data is published. However, the current implementation uses default buffering, which can result in up to 15 seconds of latency.

This PR flushes the file on each set of sentences.

It does not affect the TTY line support, although this may also be buffered.


carlesfernandez:
Thanks!

PR #706, opened 2023-01-27
do not propagate tags

jwmelto:
GNSS-SDR does not do well with tags. If they are in the data stream, downstream (acquisition and tracking) get "gummed up" as the system tries to preserve tags for insertion into the (heavily) decimated output.

However, the GNU Radio ZMQ blocks treat tags as part of the wire interface. If the source produces tags, the destination **must** enable them or the data is corrupted.

This change is specific to the ZMQ source, but any source that includes tags should similarly not pass them blindly downstream.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,7 @@ option(ENABLE_FLEXIBAND "Enable the use of the signal source adater for the Tele
 
 option(ENABLE_ARRAY "Enable the use of CTTC's antenna array front-end as signal source (experimental)" OFF)
 
-option(ENABLE_ZMQ "Enable GNU Radio ZeroMQ Messaging, requires gr-zeromq" OFF)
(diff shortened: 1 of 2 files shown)


carlesfernandez:
It's ok to have `ENABLE_ZMQ` set to `ON` by default, but the configuration fails if that component is not present. I've added a detection of the `libzmq` library, and the CMake scripts turns `ENABLE_ZMQ` to `OFF` is that library is not found. This seems to work as expected in all systems I tried, let me know if it works for you.

jwmelto:
> It's ok to have `ENABLE_ZMQ` set to `ON` by default, but the configuration fails if that component is not present. I've added a detection of the `libzmq` library, and the CMake scripts turns `ENABLE_ZMQ` to `OFF` is that library is not found. This seems to work as expected in all systems I tried, let me know if it works for you.

I didn't check, but I assumed that was how it worked, knowing the original ZMQ integration cloned existing tests. I have no issue with the fall-back logic, since ZMQ is definitely optional. Thanks!

PR #738, opened 2023-10-10
Python tools

Perrielornitorrinco:
**Python Tools**

This code includes some utilities implemented in python based on the existing utilities in matlab folder.
The results obtained during its use will be saved in a new folder called PLOTS. Inside where the data are read. (It's necessary to change the read and save paths. Similar to matlab utilities)

Location: gnss-sdr/src/utils/python
Developer: Irene Pérez Riega iperrie@inta.es

### DIFF
--- a/src/utils/python/dll_pll_veml_plot_sample.py
+++ b/src/utils/python/dll_pll_veml_plot_sample.py
@@ -1,0 +1,110 @@
+"""
+ dll_pll_veml_plot_sample.py
+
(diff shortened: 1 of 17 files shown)


carlesfernandez:
This is a very nice contribution. Could you please rebase your commit?

Perrielornitorrinco:
Thank you very much Carles! I think it is now ready to merge.

Issue #773, opened 2024-05-23
Linker error in Debug build: absl_log_internal_check_op DSO missing from command line


castle055:
I just forked the repository and was trying to setup a suitable development environment when I encountered a linker issue regarding abseil's logging library. Interestingly, this only happens when compiling the debug build.

The error in question is the following:

```
[100%] Linking CXX executable gnss-sdr
/usr/bin/ld: ../core/monitor/libcore_monitor.a(gnss_synchro.pb.cc.o): undefined reference to symbol '_ZN4absl12lts_2024011612log_internal17MakeCheckOpStringIPKvS4_EEPNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEET_T0_PKc'
/usr/bin/ld: /usr/lib/libabsl_log_internal_check_op.so.2401.0.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[3]: *** [src/main/CMakeFiles/gnss-sdr.dir/build.make:240: src/main/gnss-sdr] Error 1
make[2]: *** [CMakeFiles/Makefile2:2403: src/main/CMakeFiles/gnss-sdr.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2410: src/main/CMakeFiles/gnss-sdr.dir/rule] Error 2
make: *** [Makefile:712: gnss-sdr] Error 2
```

This seems to indicate that the `core_monitor` CMake target is not linking against `libabsl_log_internal_check_op`.

I don't have much experience with the abseil library, so I'm not sure if this is an environment issue on my part or if this is a bug in gnss-sdr's build system. However, adding `absl::log_internal_check_op` within the PRIVATE scope to the `target_link_libraries(core_monitor ...)` call in `src/core/monitor/CMakeLists.txt` seems to solve the issue:

```cmake
target_link_libraries(core_monitor
    PUBLIC
        Boost::headers
        Boost::system
        Gnuradio::runtime
        protobuf::libprotobuf
        core_system_parameters
    PRIVATE
        Boost::serialization
        absl::log_internal_check_op # <- Added this line
)
```

If this is a bug, I can submit a PR for it, but I'll need to check it doesn't conflict with any build options first because I suspect this is just a duck tape fix.

I'm compiling the `next` branch, specifically commit `f80c537` with system provided libabsl:
https://github.com/gnss-sdr/gnss-sdr/tree/f80c5373b78b5124e70d1e6743654a79145f0721

[...]

PR #774, opened 2024-05-24
Conditionally link protobuf to its dependencies in Debug mode

castle055:
Some versions of Protocol Buffers don't export their dependencies when calling `find_package(Protobuf)` from CMake, thus causing link errors from missing symbols from `abseil-cpp`.

This seems to be a known upstream bug:
https://github.com/protocolbuffers/protobuf/issues/12637

This PR implements a somewhat hacky solution shown in a comment:
https://github.com/protocolbuffers/protobuf/issues/12637#issuecomment-1871458639

This will link the results of calling `PkgConfig` for `protobuf` to the target imported by `find_package(Protobuf)`, but only if the following conditions apply:

- Build type is **Debug** (For some reason, this only happens in Debug)
- `abseil-cpp` was found
- `Protobuf` was actually found as a **system package**
- `Protobuf` version is >= 22.0 and < 26.0

This PR should resolve Issue #773

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2623,6 +2623,8 @@ else()
     find_package(Protobuf)
 
     if((CMAKE_BUILD_TYPE STREQUAL "Debug") AND Protobuf_FOUND AND absl_FOUND)
+        # This Regular Expression is used to convert the version string provided by `find_package(Protobuf)` into the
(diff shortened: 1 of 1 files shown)


carlesfernandez:
Excellent, thanks!

PR #782, opened 2024-07-02
Fixed typo in collection of A0UTC parameter for BeiDou

cl-cv:
Lines were presumably copy and pasted from other satellite system setups. A0 was still using GPS values.


carlesfernandez:
Thanks for the catch. Cherry-picked to the `next` branch.

PR #801, opened 2024-09-09
Udp ports

xguerreropau:
gnss-sdr is now able to send the same monitor UDP packages to multiple ports


carlesfernandez:
Thanks!

jwmelto:
Shouldn't this change also have updated the sample configuration files? It is a breaking change for existing users. Another option would have been to leave the keyword unchanged, since existing single port configs would still work (maybe)

Don't get me wrong, I've wanted this change for years. I'm just trying to consider existing users.

carlesfernandez:
Hi @jwmelto,

this change is not breaking any existing configuration, as far as I can see. The parameter name remains the same (`Monitor.udp_port`), and specifying a single port is still valid. Have I missed something?

jwmelto:
You are right; the variable changed but not the config tag. I didn't read carefully. Thanks.

PR #802, opened 2024-09-09
Added presets

xguerreropau:
Added presets for an easier configuration and building in cmake:
Example for a release build:

1. Configure:
$ cmake --preset config-release

2. Build:
$ cmake --build --preset build-release

The output will be generated at:
build-release/


jwmelto:
The problem with presets is that they can vary so much between installations. I wonder if this properly belongs as part of the source tree or if it just a user customization?

PR #806, opened 2024-09-25
Deleted build and moved tests and utils out from src

xguerreropau:
The build directory will be generated by the build system (cmake).
The data directory can be created by any user and execute gnss-sdr there.
The utils and tests directories don't belong into the src directory.


carlesfernandez:
I like this change, as it sorts out the source tree in a more standard way and this can help to newcomers. However, it is a breaking change for many CI setups and other kind of workflows, so we need to document it well and update every reference to that (for instance in the main README, in the docs/CHANGELOG.m file, and everywhere in the website, after fixing our own CIs and workflows. So more work is needed, but thanks for this proposal!

carlesfernandez:
Added to the `next` branch. I've included this note in the CHANGELOG:

- **Potential Breaking Change**: The source tree has been refactored to follow a
  more conventional folder structure. This may disrupt user pipelines that
  relied on the previous structure and could break development branches that
  were branched off from `next` before this change. The key changes are:

  - The `tests` and `utils` directories have been moved from the `src` folder to
    the root of the source tree.
  - The empty `build` and `data` folders have been removed. Users can create a
    building folder using `mkdir build` or by having CMake handle it:
    `cmake -S . -B build`.
  - All default names for dump or input files starting with `../data/<filename>`
    have been changed to `./<filename>`.

PR #810, opened 2024-09-30
GSOC 24: Integration of SBAS (EGNOS/WAAS) in GNSS-SDR

juanalfarom:
I have implemented the capability to perform acquisition and tracking of SBAS satellites within the GNSS-SDR framework. This feature enhances the system's ability to handle satellite-based augmentation signals, improving the overall performance and precision of the navigation solution.

As for the telemetry decoder functionality, full integration within GNSS-SDR has not yet been completed. Currently, the implementation allows the navigation bits to be saved, although some final adjustments are still required to refine their processing. In parallel, a MATLAB program, which will be uploaded in the coming weeks, enables loading all the saved navigation bits and decoding specific message types such as 0, 17, 26, and 27. Further enhancements and tests will be conducted to ensure full decoding capability and alignment with GNSS-SDR.

The following files are the main ones which have been modified during this time to achieve this implementation:

- `src/core/system_parameters/gnss_satellite.cc`: 
  - Updated to incorporate the PRN of SBAS satellites.

- `src/algorithms/telemetry_decoder/adapters/sbas_l1_telemetry_decoder.cc`: 
  - Modified to handle the decoding of SBAS L1 telemetry, enabling the system to process and store navigation bits, with further adjustments planned to improve decoding accuracy.

One last future implementation will involve creating a dedicated SBAS signal type within the GNSS-SDR framework. However, since SBAS currently operates on the same frequency as GPS L1, for the time being, the GPS L1 signal type will be used in the `.conf` file (1C). Once the SBAS-specific signal type is developed, it will be fully integrated into the system for more robust handling of SBAS signals. 

For the current changes, an example of a `.conf` file fields which have to be included to acquire 3 SBAS satellites would look like the following:

```bash
;######### CHANNELS GLOBAL CONFIG ############
Channels_1C.count=3

Channel0.satellite=121
Channel1.satellite=123
Channel2.satellite=136

;######### SBAS ACQUISITION CONFIG ############
Acquisition_1C.implementation=GPS_L1_CA_PCPS_Acquisition
Acquisition_1C.item_type=gr_complex
Acquisition_1C.coherent_integration_time_ms=1
Acquisition_1C.threshold=1.8
Acquisition_1C.use_CFAR_algorithm=false
;Acquisition_1C.blocking=true
Acquisition_1C.doppler_max=5000
Acquisition_1C.doppler_step=500	 

Acquisition_1C.dump=false
Acquisition_1C.dump_filename=./acquisition/acq_dump.dat
Acquisition_1C.dump_mat=true

;######### TRACKING SBAS CONFIG ############
Tracking_1C.implementation=GPS_L1_CA_DLL_PLL_Tracking
Tracking_1C.item_type=gr_complex
Tracking_1C.extend_correlation_symbols=1
Tracking_1C.early_late_space_chips=0.5
Tracking_1C.pll_bw_hz=35
Tracking_1C.dll_bw_hz=4
Tracking_1C.order=3
Tracking_1C.dump=true
Tracking_1C.dump_filename=./tracking/tracking_raw.dat
Tracking_1C.dump_mat=true

;######### TELEMETRY DECODER SBAS CONFIG ############
TelemetryDecoder_1C.implementation=SBAS_L1_Telemetry_Decoder
TelemetryDecoder_1C.dump=true
TelemetryDecoder_1C.dump_mat=true

### CODE REVIEW

**Review by carlesfernandez — CHANGES_REQUESTED** [review]
See comments below

**Comment by juanalfarom** [comment]
I have made a new commit applying the suggested changes, I think everything is fine now.

**Comment by carlesfernandez** [comment]
Thanks, it looks much better. Let's see how the tests go.

**Comment by juanalfarom** [comment]
New commit with clang-format and cpplink error corrections done.

File: src/algorithms/telemetry_decoder/gnuradio_blocks/gps_l1_ca_telemetry_decoder_gs.cc [hunk 1]
```diff
@@ -692,4 +692,4 @@ int gps_l1_ca_telemetry_decoder_gs::general_work(int noutput_items __attribute__
         }
 
     return 0;
-}
+}
```
jwmelto, line 695: This change should not have been committed. Please figure out how to configure your editor or IDE to ensure that all files end in a newline.

There's no other change in this file.

src/algorithms/telemetry_decoder/gnuradio_blocks/sbas_l1_telemetry_decoder_gs.cc
```diff
@@ -41,20 +42,23 @@
 
 sbas_l1_telemetry_decoder_gs_sptr sbas_l1_make_telemetry_decoder_gs(
     const Gnss_Satellite &satellite,
-    bool dump)
+    const Tlm_Conf &conf)
 {
-    return sbas_l1_telemetry_decoder_gs_sptr(new sbas_l1_telemetry_decoder_gs(satellite, dump));
+    return sbas_l1_telemetry_decoder_gs_sptr(new sbas_l1_telemetry_decoder_gs(satellite, conf));
```
jwmelto, line 47: The body of legacy examples notwithstanding, in modern C++ explicit calls to `operator new` generally indicate a memory management problem. Modern code should never have `new` or `delete`; use the appropriate type of managing pointer and its associated factory function. Here, consider
```suggestion
    return std::make_shared<sbas_l1_telemetry_decoder_gs>(satellite, conf);
```

Note that @carlesfernandez may have a different opinion, but I don't see a viable path for adding enhancements and supporting archaic environments. This is just my opinion, but I would write all **new** features for current and future platforms/environments, not frozen legacy ones. 

The key is not to break anything in the process, so it may be that new features need to be conditionally enabled so obsolete environments can ignore them. Personally, I'd freeze a release that supports the old environments that we care about and require building from source (HEAD) to be up-to-date for standards and external versions. That's outside the scope of your change, so you don't need to fix it, but you do need to be aware of the tension.

src/algorithms/telemetry_decoder/gnuradio_blocks/sbas_l1_telemetry_decoder_gs.cc
```diff
@@ -41,20 +42,23 @@
...
-                 d_channel(0),
-                 d_block_size(D_SAMPLES_PER_SYMBOL * D_SYMBOLS_PER_BIT * D_BLOCK_SIZE_IN_BITS)
+    const Tlm_Conf &conf) : gr::block("sbas_l1_telemetry_decoder_gs",
+                                gr::io_signature::make(1, 1, sizeof(Gnss_Synchro)),
+                                gr::io_signature::make(1, 1, sizeof(Gnss_Synchro))),
+                            d_dump(true),
+                            d_dump_mat(true),
+                            d_remove_dat(true),
+                            d_dump_filename("tele"),
+                            d_channel(0),
```
jwmelto, line 60: I don't think these should be hard-coded. You created a config object; hopefully it encompasses these values. Use the config here.

Note that if (in particular) `d_channel` is hard-coded to 0, why bother keeping the state in this object?

src/algorithms/telemetry_decoder/gnuradio_blocks/sbas_l1_telemetry_decoder_gs.cc
```diff
@@ -82,6 +88,24 @@ sbas_l1_telemetry_decoder_gs::~sbas_l1_telemetry_decoder_gs()
                 {
                     LOG(WARNING) << "Exception in destructor closing the dump file " << ex.what();
                 }
+            if (pos == 0)
+                {
+                    if (!tlm_remove_file(d_dump_filename))
+                        {
+                            LOG(WARNING) << "Error deleting temporary file";
+                        }
```
jwmelto, line 96: Is this necessary or desirable? If a dump file is configured (!!) then it should be produced, even if it is 0 bytes in length. It would be confusing to look for your dump file and not find it. If this is consistent with the rest of the code base, never mind. Consistency can be more important than correctness or rationality, when it comes to maintenance.

src/algorithms/telemetry_decoder/gnuradio_blocks/sbas_l1_telemetry_decoder_gs.cc
```diff
@@ -96,7 +120,33 @@ void sbas_l1_telemetry_decoder_gs::set_satellite(const Gnss_Satellite &satellite
 void sbas_l1_telemetry_decoder_gs::set_channel(int32_t channel)
 {
     d_channel = channel;
-    LOG(INFO) << "SBAS channel set to " << channel;
+    DLOG(INFO) << "Navigation channel set to " << channel;
+    // ############# ENABLE DATA FILE LOG #################
```
jwmelto, line 124: You've created an interesting dependency here. Now, this function is responsible for opening (truncating) the dump file. If this function is never called, will the dump file be created elsewhere?

If the dump file was previously opened and then this function is called, the internal state describing the channel will be updated but the dump-file will continue to be the old channel. That can't be good. It seems that setting the channel and opening (re-opening?) the dump file are separate concerns. Perhaps they shouldn't be mixed here.

src/algorithms/telemetry_decoder/gnuradio_blocks/sbas_l1_telemetry_decoder_gs.cc
```diff
@@ -415,7 +465,6 @@ int sbas_l1_telemetry_decoder_gs::general_work(int noutput_items __attribute__((
     current_symbol = in[0];
     // copy correlation samples into samples vector
     d_sample_buf.push_back(current_symbol.Prompt_I);  // add new symbol to the symbol queue
-
```
jwmelto, line 418: don't hate the vertical whitespace. In particular, blank lines before comments facilitate readability. Beware of gratuitously changing things to match **your** IDE; your change may not aid readability on someone else's tool.

src/algorithms/telemetry_decoder/gnuradio_blocks/sbas_l1_telemetry_decoder_gs.cc
```diff
@@ -473,6 +522,24 @@ int sbas_l1_telemetry_decoder_gs::general_work(int noutput_items __attribute__((
             d_sample_buf.clear();
         }
 
+    if (d_dump == true)
+        {
+            // MULTIPLEXED FILE RECORDING - Record results to file
+            try
+                {
+                    double tmp_double;
+                    uint64_t tmp_ulong_int;
```
jwmelto, line 531: unused. Didn't you get a compiler warning?

src/algorithms/telemetry_decoder/libs/tlm_conf.cc
jwmelto, line 1: this file is a gratuitous whitespace change. I recommend reverting it and not modifying it for this PR

src/algorithms/tracking/gnuradio_blocks/gps_l1_ca_dll_pll_tracking_gpu_cc.cc
jwmelto, line 1: this file is a gratuitous (negative) whitespace change. I recommend reverting it and not modifying it for this PR

src/core/system_parameters/gnss_satellite.cc
```diff
@@ -183,7 +182,7 @@ void Gnss_Satellite::set_PRN(uint32_t PRN_)
         }
     if (system == "GPS")
         {
-            if (PRN_ < 1 or PRN_ > 32)
+            if (PRN_ < 1 or (PRN_ > 32 and PRN_ < 120))
```
jwmelto, line 185: Consider writing your 3-way compares consistently with those languages that support that expression
```suggestion
            if (PRN_ < 1 or (32 <  PRN_ and PRN_ < 120))
```

jwmelto: While you described your overloading of GPS (and I don't think you originated it) for SBAS, this is a place where it would be worth making the distinction.

BY DEFINITION, the GPS constellation will not be more than 32 satellites. The higher numbered satellites are not (really) GPS satellites, are they? So is this a valid GPS error bound anymore?

src/core/system_parameters/gnss_satellite.cc
```diff
@@ -405,6 +404,24 @@ std::string Gnss_Satellite::what_block(const std::string& system_, uint32_t PRN_
                 case 32:
                     block_ = std::string("IIF");  // Plane F
                     break;
+                case 120:
```
jwmelto, line 407: What you did here is consistent with the legacy code (good) but I don't really love it. This validation is begging for a database of some sort, even if it's just a map of satellites per constellation that is either initialized from a file or in an initializer list. This 400 lines of code is unwarranted complexity. 

The question of when to refactor is a subjective one, and it best comes with experience. You did well to follow the pattern, but I hope you had to hold your nose while you did it.

jwmelto (commented):
It looks like 90% of this code change was replacing the simple `bool` "dump" parameter with a config structure, but I don't see a lot of meat.

The write-up says you "implemented the capability to perform acquisition and tracking of SBAS satellites" but I'm not really seeing that, unless the real magic was just adding them to **gnss_satellite.cc**. Did you maybe remove too much in trying to address previous comments?

juanalfarom:
After obtaining the cpplint error in so many files which were no modified by me, I have decided to include the files changes during the project on a fresh folder of GNSS-SDR-next. That is the reason of having now so many modified files, but they are the same as the current ones in GNSS-SDR-next.

PR #839, opened 2025-01-08
Gps almanac support

joebre:
Implementation of almanac decoding for GPS. 
The correctness of the decoded almanacs has been verified by comparing the GNSS-SDR generated almanac XML with the YUMA files provided @  http://celestrak.org/GPS/almanac/Yuma/
The almanac for a particular PRN gets published once both the almanac data and almanac week number have been decoded.


jwmelto:
Please make this PR against the **next** branch, not **main**.

PR #866, opened 2025-03-24
Fix: Display clear HackRF One channel limit error

Ceres2805:
This commit improves the error message displayed when users
incorrectly configure GNSS-SDR to use more than one channel with
the HackRF One. The previous message was vague, leading to user
confusion.

The new error message explicitly states the single-channel
limitation and directs users to the documentation for correct
configuration examples. This prevents users from spending time
debugging an invalid configuration.

Reference issue: [#803](https://github.com/gnss-sdr/gnss-sdr/issues/803)

Issue #803, opened 2024-09-19
I encountered a new issue.Can someone help me understand why it's happening?


pyjhky1315:
**### _it is the err:_**
eugene@ubuntu:~/gnss-sdr/conf$ gnss-sdr --config_file=./hackrf.conf
linux; GNU C++ version 7.3.0; Boost_106501; UHD_003.010.003.000-0-unknown

Initializing GNSS-SDR v0.0.19.git-main-87fcfd237 ... Please wait.
Logging will be written at "/tmp"
Use gnss-sdr --log_dir=/path/to/log to change that.
OsmoSdr arguments: hackrf=0,bias=1,buffers=256
gr-osmosdr 0.1.4 (0.1.4) gnuradio 3.7.11
built-in source types: file osmosdr fcd rtl rtl_tcp uhd miri hackrf bladerf rfspace airspy airspyhf soapy redpitaya freesrp
Using HackRF One with firmware local-79baef7
Using 256 buffers of size 262144.
Enabled antenna bias voltage
Actual RX Rate: 2e+07 [SPS]...
Actual RX Freq: 1.56742e+09 [Hz]...
PLL Frequency tune error: 0 [Hz]...
Actual RX Gain: 0 dB...
Actual Bandwidth: 1.5e+07 [Hz]...
RF Channels: 2
Starting a TCP/IP server of RTCM messages on port 2101
The TCP/IP server of RTCM messages is up and running. Accepting connections ...
gnss-sdr: /usr/include/boost/smart_ptr/shared_ptr.hpp:734: typename boost::detail::sp_member_access<T>::type boost::shared_ptr<T>::operator->() const [with T = gr::basic_block; typename boost::detail::sp_member_access<T>::type = gr::basic_block*]: Assertion `px != 0' failed.
Aborted (core dumped)
eugene@ubuntu:~/gnss-sdr/conf$
eugene@ubuntu:~/gnss-sdr/conf$

**it is the **confile****
eugene@ubuntu:~/gnss-sdr/conf$ cat hackrf.conf
; This is a GNSS-SDR configuration file
; The configuration API is described at https://gnss-sdr.org/docs/sp-blocks/
; SPDX-License-Identifier: GPL-3.0-or-later
; SPDX-FileCopyrightText: (C) 2010-2020  (see AUTHORS file for a list of contributors)

; You can define your own receiver and invoke it by doing
; gnss-sdr --config_file=my_GNSS_SDR_configuration.conf
;

[...]

PR #867, opened 2025-03-24
Fix: Display clear HackRF One channel limit error

Ceres2805:
This commit improves the error message displayed when users
incorrectly configure GNSS-SDR to use more than one channel with
the HackRF One. The previous message was vague, leading to user
confusion.

The new error message explicitly states the single-channel
limitation and directs users to the documentation for correct
configuration examples. This prevents users from spending time
debugging an invalid configuration.

Reference issue: [#803](https://github.com/gnss-sdr/gnss-sdr/pull/866)

### DIFF
--- a/.github/workflows/citation.yml
+++ b/.github/workflows/citation.yml
@@ -24,7 +24,7 @@ jobs:
         uses: dieghernan/cff-validator@main
       # Upload artifact in case of failure
       - name: Update artifact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: failure()
         with:
           name: citation-cff-errors
--- a/.github/workflows/gnss-sdr_archs.yml
+++ b/.github/workflows/gnss-sdr_archs.yml
@@ -41,7 +41,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: uraimo/run-on-arch-action@v2.6.0
+      - uses: uraimo/run-on-arch-action@v2.7.1
         name: Test in non-x86 container
         continue-on-error: ${{ contains(fromJson('["ppc64le", "s390x"]'), matrix.arch) }}
         id: test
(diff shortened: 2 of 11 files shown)


Teja2007-max:
This thread is about a pull request to **GNSS-SDR** that improves the error message when users configure multiple channels with a **HackRF One**, which only supports a single RF channel.

Now the concern raised is:

> “It is possible that your PR breaks some configurations. Could you test this config file…”

The referenced configuration:
`gnss-sdr_multichannel_GPS_Galileo_Beidou_L1_hackrf_realtime.conf`

---

## What’s Actually Being Questioned?

There’s an important distinction:

* **RF channels (hardware channels)** → HackRF One supports only **one**
* **GNSS tracking channels (software channels)** → GNSS-SDR can track many satellites simultaneously using a single RF stream

If the new error message checks the wrong parameter (e.g., number of tracking channels instead of number of RF sources), it could incorrectly reject valid multichannel GNSS configs.

That’s likely what the reviewer is worried about.

---

## What Should Be Tested

To ensure the PR doesn’t break valid setups:

1. Test the linked config file as-is.
2. Verify:

   * It uses only **one RF source**
   * It allows multiple tracking channels
3. Confirm that:

   * No error is thrown for valid multichannel tracking
   * The error triggers only if multiple RF inputs are defined

---

## What Could Break

If the new validation logic checks something like:

```
if (channels > 1)
    error
```

instead of:

```
if (rf_sources > 1)
    error
```

then valid configurations (GPS + Galileo + BeiDou tracking) would fail incorrectly.

---

## Suggested Reply (Professional & Clear)

You might respond with something like:

> Thanks for pointing that out.
> I will test the provided multichannel HackRF configuration to verify that the check only restricts multiple RF sources and does not interfere with valid multi-satellite tracking setups.
>
> The intention of this PR was to prevent invalid multi-RF configurations, not to limit tracking channels. I’ll confirm behavior and report back.

---

If you’d like, you can paste the relevant validation code here and I can help analyze whether it might incorrectly reject valid configs.

PR #884, opened 2025-05-06
Cleanup gnss flowgraph available prns

MathieuFavreau:
Refactor to remove some code duplication within `gnss_flowgraph.cc`.
Move the available PRN lists to a `std::unordered_map` to remove duplicating code for each signal.

The motivation behind this change is to simplify supporting new signals, and to simplify fixing bugs within this file.
Another pull request is coming after this one to fix a bug where a satellite can be tracked multiple times (I want to keep both PRs separated to make the bug fix easier to review).


carlesfernandez:
Thanks, I love this PR. I've made a small change to support C++ standards older than 17 (which didn't support structured bindings) but, if the available standard is high enough, the code remains as you proposed.

PR #886, opened 2025-05-07
Cleanup gnss block factory

MathieuFavreau:
Updating `gnss_block_factory.cc` is error prone since you have to add your implementations in two places for Acquisition, Tracking and Telemetry, with this change it is only required to do it once, removing code duplication.

Also I do not see a point in creating a temporary `block_` variable to then assign it to `block`, and that change alone removes hundreds of lines of code.

Finally `std::unique_ptr<std::vector<...>>` doesn't make sense, `std::vector<...>` is already using an internal pointer, this is just unnecessary indirection.

### DIFF
--- a/src/core/receiver/gnss_block_factory.cc
+++ b/src/core/receiver/gnss_block_factory.cc
@@ -240,6 +240,501 @@ auto findRole(ConfigurationInterface const* configuration, std::string const& ba
         }
     return role;
 };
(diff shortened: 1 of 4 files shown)


carlesfernandez:
Excellent! I really appreciate this work, thank you @MathieuFavreau! Merging.

PR #887, opened 2025-05-08
Cleanup receiver type

MathieuFavreau:
These super long `if` statements were getting hard to maintain, the proof is that receiver types 601, 602 and 603 did not match their definition:

601
- Definition: BeiDou B3I + GPS L2C
- Actual: BeiDou B3I + GPS L2C + GPS L1 C/A

602
- Definition: Beidou B3I + GLONASS L2 C/A
- Actual: Beidou B3I + GLONASS L2 C/A + GPS L1 C/A

603
- Definition: Beidou B3I + GPS L2C + GLONASS L2 C/A
- Actual: Beidou B3I + GPS L2C + GLONASS L2 C/A + GPS L1 C/A

I've decided to change to match the definition, but we could also change the definition if we don't want to introduce a change.
Also the definition comment included values 604 to 610, but they were not implemented so I removed them from the comment.


carlesfernandez:
Excellent again!

I moved receiver_type{.cc,.h} from PVT/adapters to PVT/libs for consistency.
Tweaked a bit to make it build in old compilers, while still making use of modern features if available.

Thanks, merging.

PR #892, opened 2025-05-28
Ensure correct file length and sample skip calculations on 32-bit systems by using uint64_t

mmajoral:
Ensure correct file length and sample skip calculations on 32-bit systems by using uint64_t
    
On 32-bit architectures, size_t may not be large enough to represent large file sizes or sample counts. This update modifies the file length computation and the return type of samplesToSkip() to use uint64_t, allowing accurate handling of large files and ensuring consistency across platforms.

### DIFF
--- a/src/algorithms/signal_source/adapters/file_source_base.cc
+++ b/src/algorithms/signal_source/adapters/file_source_base.cc
@@ -365,16 +365,16 @@ std::tuple<size_t, bool> FileSourceBase::itemTypeToSize()
 double FileSourceBase::packetsPerSample() const { return 1.0; }
 
 
-size_t FileSourceBase::samplesToSkip() const
+uint64_t FileSourceBase::samplesToSkip() const
 {
-    auto samples_to_skip = size_t(0);
+    auto samples_to_skip = uint64_t(0);
 
     if (seconds_to_skip_ > 0)
         {
             // sampling_frequency is in terms of actual samples (output packets). If this source is
             // compressed, there may be multiple packets per file (read) sample. First compute the
             // actual number of samples to skip (function of time and sample rate)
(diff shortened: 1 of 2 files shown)


jwmelto:
Is there a failing test that this fixes? [`std::size_t`](https://en.cppreference.com/w/cpp/types/size_t.html) is guaranteed to be large enough to represent any possible size. In particular,

> If a program attempts to form an oversized type (i.e., the number of bytes in its [object representation](https://en.cppreference.com/w/cpp/language/objects.html#Object_representation_and_value_representation) exceeds the maximum value representable in std::size_t), the program is ill-formed.

mmajoral:
Hi @jwmelto ,
I cross-compiled GNSS-SDR for the 32-bit ARM architecture on the Zynq-7000 platform, and ran it in post-processing mode. However, I encountered this issue: on the 32-bit platform, size_t appears to limit the maximum file size to 32 bits, causing GNSS-SDR to stop processing the input file earlier than expected.

I found some references, for example: [https://stackoverflow.com/questions/918787/whats-sizeofsize-t-on-32-bit-vs-the-various-64-bit-data-models](https://stackoverflow.com/questions/918787/whats-sizeofsize-t-on-32-bit-vs-the-various-64-bit-data-models), which indicate that the C++ standard allows implementations to choose the representation of size_t, with range up to SIZE_MAX, which must be at least 65,535.

To verify this, I printed the size of size_t and the value of SIZE_MAX on both platforms:

PC (x86-64):
    sizeof(size_t): 8 bytes
    SIZE_MAX: 18446744073709551615

ARM (32-bit):
    sizeof(size_t): 4 bytes
    SIZE_MAX: 4294967295

Replacing size_t with uint64_t solved this issue.

PR #894, opened 2025-06-09
Integrating External Sensor Measurements into the GNSS-SDR flowgraph

castle055:
This PR is a part of my project for this year's Google Summer of Code, see [GNSS + INS Sensor Fusion](https://summerofcode.withgoogle.com/programs/2025/projects/ItPRA36a).

This proposal offers a new feature for the purpose of integrating external sensor measurements into the RF sample stream, making them available to any block in the GNSS-SDR flowgraph. It can read sensor measurements from one or multiple files and synchronize them with the RF sample stream. It is highly configurable and allows for a wide range of file formats.

This PR also includes a simple proof of concept. The Kalman filter in the PVT block can be configured to use velocity measurements from an IMU in its estimations. This is disabled by default and depends on `PVT.enable_pvt_kf` being `true` in the configuration file. More information in the _Usage_ section.

For now it only works with `FileSignalSource` but future development could integrate it with other signal sources, or even connect it directly in the flowgraph.

# Usage

From the user's perspective, some GNSS-SDR algorithms (such as the Kalman filter in the PVT block) may be configured to work with sensor data (`PVT.kf_use_imu_vel = true`). Only in such cases, GNSS-SDR will require certain sensors to be provided. If they are not provided, an exception is thrown since the requested algorithm cannot work without them.

In the Kalman filter example, the required measurements are the velocities in the three axis. These can be provided by adding the following `SensorData` section to the configuration file:

```
;######### SENSOR DATA CONFIG ############  
SensorData.enabled=true  

; Configure data file
SensorData.file0.filename=/path/to/sensor/data/file.dat  
SensorData.file0.repeat=true  
SensorData.file0.chunk_size=12    ; (3 floats 4 bytes wide each)
SensorData.file0.file_offset=0
SensorData.file0.sample_offset=0
SensorData.file0.sample_period=300000

; Configure sensors
SensorData.sensor0.data=IMU_VEL_X
SensorData.sensor0.type=f32       ; The values are stored as a 4 byte float
SensorData.sensor0.file=0         ; The values are in file 0
SensorData.sensor0.offset=0       ; The values are at the start of each chunk

SensorData.sensor1.data=IMU_VEL_Y  
SensorData.sensor1.type=f32
SensorData.sensor1.file=0
SensorData.sensor1.offset=4       ; values found 4 bytes from the start of each chunk

SensorData.sensor2.data=IMU_VEL_Z
SensorData.sensor2.type=f32
SensorData.sensor2.file=0
SensorData.sensor2.offset=8
```

This can be simplified taking into account the following behavior:
- All sensors except the first one default their **data type** and **file ID** to those of the previous sensor. The first sensor defaults to a **file ID** equal to 0.
- The first sensor's **offset** defaults to 0. All other sensors default to the result of adding the offset of the previous sensor and its size.

We can thus write this, having the same effect:

```
...

SensorData.sensor0.data=IMU_VEL_X
SensorData.sensor0.type=f32
SensorData.sensor1.data=IMU_VEL_Y  
SensorData.sensor2.data=IMU_VEL_Z
```

Sensor ordering is determined by their IDs, which must be unique and contiguous starting from 0. File IDs have these same requirements, but there are no automatic defaults. All fields must be specified for each file.

Finally, we can tell the PVT block to use these values in the Kalman filter:

```
;######### PVT CONFIG ############  
...

PVT.enable_pvt_kf=true
PVT.kf_use_imu_vel=true
```

## Sensor data types

Sensor measurements are propagated throughout the flowgraph using the polymorphic type `pmt::pmt_t`. This allows for sensors to work with whatever type best represents their measurements. 

Each supported sensor has a specified internal data type. GNSS-SDR will convert sensor measurements from their in-file binary representation to their respective internal type. This means neither the user nor the implementer need to worry about type conversion. The user specifies which type the sensor uses in the sensor data file. Meanwhile, the implementer will always receive the values as converted to their internal types. For instance, the user can specify inertial measurements of velocity to be read from the sensor data file using any scalar type (F32, F64, I32 or I64), but the implementer will always receive them as F32.

Internal types are specified in [SensorIdentifier\:\:get_internal_type()](https://github.com/castle055/gnss-sdr/blob/ft/sensor-data/src/algorithms/libs/sensor_data/sensor_identifier.cc#L155-L178). Types are converted using the conversion table at [conversion_table](https://github.com/castle055/gnss-sdr/blob/ft/sensor-data/src/algorithms/libs/sensor_data/sensor_identifier.cc#L30). Modification of supported types or assigned internal types implies modification of this table. An exception is thrown if the required conversion is not found.

# Details and Further Development
## Using Sensor Measurements

From the implementer's perspective, any GNSS-SDR block implementation after the signal source can access sensor measurements by making use of the class `SensorDataAggregator`.

This is done inside the GNU Radio block implementation. A `SensorDataAggregator` is added as a member field to the class. It gets initialized in the constructor as follows:

```c++
// `configuration` is a reference to an instance of `ConfigurationInterface`
SensorDataSourceConfiguration sensor_config{configuration};

// Conditionally require certain sensors
std::vector<SensorIdentifier::value_type> required_sensors{};
if (conf_.kf_use_imu_vel)
    {
        required_sensors.emplace_back(SensorIdentifier::IMU_VEL_X);
        required_sensors.emplace_back(SensorIdentifier::IMU_VEL_Y);
        required_sensors.emplace_back(SensorIdentifier::IMU_VEL_Z);
    }

// Initialize aggregator
d_sensor_data_aggregator = std::make_unique<SensorDataAggregator>(sensor_config, required_sensors);
```

`SensorDataAggregator` validates the required sensor list against the provided configuration during construction. If a required sensor is not provided, an exception is thrown with an appropriate error message. 

Then, in the `work()` function:

```c++
// Get relevant stream tags
std::vector<gr::tag_t> sensor_tags;
this->get_tags_in_range(
  sensor_tags,
  0,
  this->nitems_read(0),
  this->nitems_read(0) + noutput_items,
  d_sensor_data_aggregator->SENSOR_DATA_TAG
);
// Update the aggregator
d_sensor_data_aggregator->update(sensor_tags);

// Get sensor measurements
SensorDataSample<float> vx = d_sensor_data_aggregator.get_last_f32(SensorIdentifier::IMU_VEL_X);
vx.rf_sample_stamp; // Measurement was taken at the same time as this RF sample
vx.value;           // Measurement value

// It is possible to iterate over all new samples in the last update.
// This will include the last sample from the previous update as well, 
//  so this list will never be empty after the first sensor sample.
for (const auto& vx : sensor_data_aggregator.get_f32(SensorIdentifier::IMU_VEL_X)) {
  vx.rf_sample_stamp;
  vx.value;
}
```

## Adding more sensors

New algorithms may require measurements from sensors other than those already implemented (Barometers, LiDAR, cameras, etc...). New sensors can be added as follows:

1. Add the new identifier in `SensorIdentifier::value_type`.
2. Add the corresponding cases in the following functions:
	- `SensorIdentifier::from_string`
	- `SensorIdentifier::to_string`
	- `SensorIdentifier::get_internal_type`
3. If required, add new entries to the conversion table in `sensor_identifier.cc`

## Adding more sensor data types

Currently, only 5 data types are supported for sensor measurements (UINT64, F32, F64, I32 and I64, all little endian). It is rare for a sensor to measure unsigned values but the UINT64 type is needed internally in order to handle sample stamps.

These 5 types should cover most use cases where the sensor measurements are real numbers. However, new types can be added if needed (complex values, images, ...).

1. Add the new type in `SensorDataType::value_type`.
2. Add the corresponding cases in the following functions:
	- `SensorDataType::from_string`
	- `SensorDataType::to_string`
	- `SensorDataType::get_size`
	- `SensorDataType::make_value`
3. If required, add new entries to the conversion table in `sensor_identifier.cc`
4. The `SensorDataAggregator` class holds a map for each data type used as an internal type. If a new type is added and it is intended to be used as the internal representation for a sensor, a new map must be added and properly managed.

> Sorry for the long PR text, but this introduces some powerful functionality that is intended to be expanded upon. So I thought documenting it was important.


src/algorithms/signal_source/adapters/file_source_base.cc
```diff
@@ -205,6 +207,14 @@ void FileSourceBase::connect(gr::top_block_sptr top_block)
             DLOG(INFO) << "connected output to file sink";
         }
 
+    // EXTRA DATA
+    if (sensor_data_source())
+        {
+            top_block->connect(std::move(output), 0, sensor_data_source(), 0);
```
carlesfernandez: Please do not move output here, you are using it later on.

src/algorithms/signal_source/adapters/file_source_base.cc
```diff
@@ -253,13 +263,21 @@ void FileSourceBase::disconnect(gr::top_block_sptr top_block)
             DLOG(INFO) << "disconnected output to file sink";
         }
 
+    // EXTRA DATA
+    if (sensor_data_source())
+        {
+            // TODO - FIXME: This is NOT okay, `output` is the extra data source, not the valve/source/throttle/whatever is left of this
+            top_block->disconnect(std::move(output), 0, sensor_data_source(), 0);
```
carlesfernandez: Do not move

src/algorithms/libs/sensor_data/sensor_data_type.cc
```diff
@@ -0,0 +1,108 @@
...
+        case SensorDataType::UINT64:
+            return sizeof(uint64_t);
+        case SensorDataType::F32:
+            return sizeof(float);
+        case SensorDataType::F64:
+            return sizeof(double);
+        case SensorDataType::I32:
+            return sizeof(int);
+        case SensorDataType::I64:
+            return sizeof(long);
```
carlesfernandez: Please replace `long` by `int64_t` to ensure 64 bits in all archs.

src/algorithms/libs/sensor_data/sensor_data_type.cc
```diff
@@ -0,0 +1,108 @@
...
+        case SensorDataType::UINT64:
+            return pmt::from_uint64(*static_cast<uint64_t*>(value));
+        case SensorDataType::F32:
+            return pmt::from_float(*static_cast<float*>(value));
+        case SensorDataType::F64:
+            return pmt::from_double(*static_cast<double*>(value));
+        case SensorDataType::I32:
+            return pmt::from_long(long{*static_cast<int*>(value)});
+        case SensorDataType::I64:
+            return pmt::from_long(*static_cast<long*>(value));
```
carlesfernandez: Please replace `long` by `int64_t` to ensure 64 bits in all archs.

src/algorithms/libs/sensor_data/sensor_data_type.cc
```diff
@@ -0,0 +1,108 @@
...
+    switch (v)
+        {
+        case SensorDataType::UINT64:
+            return pmt::from_uint64(*static_cast<uint64_t*>(value));
+        case SensorDataType::F32:
+            return pmt::from_float(*static_cast<float*>(value));
+        case SensorDataType::F64:
+            return pmt::from_double(*static_cast<double*>(value));
+        case SensorDataType::I32:
+            return pmt::from_long(long{*static_cast<int*>(value)});
```
carlesfernandez: Please replace `long` by `int64_t` to ensure 64 bits in all archs.

src/algorithms/libs/sensor_data/sensor_data_source_configuration.cc
```diff
@@ -0,0 +1,216 @@
...
+        }
+    return ok;
+}
+
+bool SensorDataSourceConfiguration::validate_sensors() const
+{
+    bool ok = true;
+    for (const auto& sensor : sensors_)
+        {
+            if (not files_.contains(sensor.file_id))
```
carlesfernandez: `contains()` is only available since C++20. You can use a guard here or just use:

```
if (files_.find(sensor.file_id) == files_.end())
```

src/algorithms/libs/sensor_data/sensor_data_aggregator.cc
```diff
@@ -0,0 +1,147 @@
...
+
+            std::string key_str = pmt::write_string(key);
+            SensorIdentifier::value_type sensor_id = SensorIdentifier::from_string(key_str);
+
+            if (sensor_id != SensorIdentifier::SAMPLE_STAMP and sensor_id != SensorIdentifier::CHUNK_COUNT)
+                {
+                    switch (SensorIdentifier::get_internal_type(sensor_id))
+                        {
+                        case SensorDataType::F32:
+                            if (f32_data_.contains(sensor_id))
```
carlesfernandez: `contains()` is only available since C++20. You can use a guard here or just use:

```
if (f32_data_.find(sensor_id) != f32_data_.end())
```

src/algorithms/libs/sensor_data/sensor_data_source.cc
```diff
@@ -0,0 +1,122 @@
...
+            std::size_t s_period = file.sample_period;
+            if (items_per_sample_ != 1)
+                {
+                    s_offset *= items_per_sample_;
+                    s_period *= items_per_sample_;
+                }
+
+            sensor_data_files_.emplace(id, std::make_shared<SensorDataFile>(file.filename, s_offset, s_period, file.file_offset, file.chunk_size, file.repeat));
+
+            if (not sensor_config_map_.contains(id))
```
carlesfernandez: contains() is only available since C++20. You can use a guard here or just use:

```
if(sensor_config_map_.find(id) == sensor_config_map_.end())
```

src/algorithms/libs/sensor_data/sensor_data_aggregator.h
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * -----------------------------------------------------------------------------
+ */
+
+
+#ifndef GNSS_SDR_SENSOR_DATA_AGGREGATOR_H
+#define GNSS_SDR_SENSOR_DATA_AGGREGATOR_H
+
+#include "sensor_data_source_configuration.h"
+#include "sensor_identifier.h"
```
carlesfernandez, line 22: Please add:

```
#include <gnuradio/tags.h>
```

src/algorithms/signal_source/adapters/file_source_base.cc
```diff
@@ -574,6 +594,22 @@ gr::blocks::file_sink::sptr FileSourceBase::create_sink()
...
 
+SensorDataSource::sptr FileSourceBase::create_sensor_data_source()
+{
+    if (sensor_data_source_configuration_.is_enabled())
+        {
+            if (is_complex_)
+                {
+                    sensor_data_source_configuration_.set_items_per_sample(2);
+                }
+            sensor_data_source_ = std::make_shared<SensorDataSource>(
```
carlesfernandez: please use `gnss_make_shared` instead of `std::make_shared` for backward compatibility:

```
sensor_data_source_ = gnss_make_shared<SensorDataSource>(
...
```

src/algorithms/libs/sensor_data/sensor_data_file.h
```diff
@@ -0,0 +1,80 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * -----------------------------------------------------------------------------
+ */
+
+
+#ifndef GNSS_SDR_SENSOR_DATA_FILE_H
+#define GNSS_SDR_SENSOR_DATA_FILE_H
+
+#include <cstddef>  // for size_t
```
carlesfernandez, line 21: Please add
```
#include <gnss_block_interface.h>
```

src/algorithms/libs/sensor_data/sensor_data_file.h
```diff
@@ -0,0 +1,80 @@
...
+/** \addtogroup Algorithms_Library
+ * \{ */
+/** \addtogroup Algorithm_libs algorithms_libs
+ * \{ */
+
+
+class SensorDataFile
+{
+public:
+    using sptr = std::shared_ptr<SensorDataFile>;
```
carlesfernandez: Please use:

```
using sptr = gnss_shared_ptr<SensorDataFile>;
```

src/algorithms/libs/sensor_data/sensor_data_file.h
```diff
@@ -0,0 +1,80 @@
...
+private:
+    bool read_item(std::vector<uint8_t>& buffer);
+
+    void read_into_io_buffer();
+
+    void read_into_item_buffer(std::vector<uint8_t>& item_buf);
+
+    std::string path_;
+    std::ifstream file_;
+    std::size_t sample_delay_;
```
carlesfernandez: member `sample_delay_` is set but not used

src/algorithms/libs/sensor_data/sensor_data_file.cc
```diff
@@ -0,0 +1,126 @@
...
+SensorDataFile::SensorDataFile(
+    const std::string& path,
+    const std::size_t& sample_delay,
+    const std::size_t& sample_period,
+    const std::size_t& offset_in_file,
+    const std::size_t& item_size,
+    const bool& repeat)
+    : path_(path),
+      file_(path_, std::ios::binary),
+      sample_delay_(sample_delay),
```
carlesfernandez: `sample_delay_` is set but not used.

src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc
```diff
@@ -657,6 +657,22 @@ void hybrid_observables_gs::set_tag_timestamp_in_sdr_timeframe(const std::vector
         }
 }
 
+void hybrid_observables_gs::propagate_sensor_data(const std::vector<Gnss_Synchro> &data)
```
carlesfernandez: Why passing `data` if it is not used?

src/algorithms/signal_source/adapters/file_source_base.cc
```diff
@@ -253,13 +263,21 @@ void FileSourceBase::disconnect(gr::top_block_sptr top_block)
             DLOG(INFO) << "disconnected output to file sink";
         }
 
+    // EXTRA DATA
+    if (sensor_data_source())
+        {
+            // TODO - FIXME: This is NOT okay, `output` is the extra data source, not the valve/source/throttle/whatever is left of this
+            top_block->disconnect(std::move(output), 0, sensor_data_source(), 0);
+            DLOG(INFO) << "disconnected output to extra data source";
+            output = sensor_data_source();
```
carlesfernandez, line 271: why?

src/algorithms/signal_source/adapters/file_source_base.cc
```diff
@@ -205,6 +207,14 @@ void FileSourceBase::connect(gr::top_block_sptr top_block)
             DLOG(INFO) << "connected output to file sink";
         }
 
+    // EXTRA DATA
+    if (sensor_data_source())
+        {
+            top_block->connect(std::move(output), 0, sensor_data_source(), 0);
+            DLOG(INFO) << "connected output to sensor data source, which now becomes the new output";
+            output = sensor_data_source();
```
carlesfernandez, line 215: why?

src/algorithms/libs/sensor_data/sensor_identifier.cc
```diff
@@ -0,0 +1,190 @@
...
+            if (conversion_entry.from == from_type && conversion_entry.to == to_type)
+                {
+                    return &conversion_entry;
+                }
+        }
+    return nullptr;
+}
+
+
+SensorIdentifier::value_type SensorIdentifier::from_string(const std::string& s)
```
carlesfernandez, line 50: Maybe keeping these equivalences in an unordered_map would be easier to manipulate and to maintain?

src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
```diff
@@ -1736,6 +1736,9 @@ int64_t dll_pll_veml_tracking::uint64diff(uint64_t first, uint64_t second)
 int dll_pll_veml_tracking::general_work(int noutput_items __attribute__((unused)), gr_vector_int &ninput_items,
     gr_vector_const_void_star &input_items, gr_vector_void_star &output_items)
 {
+    // std::vector<gr::tag_t> tags{};
+    // get_tags_in_window(tags, 0, 0, ninput_items[0], pmt::mp("sensor_data"));
```
carlesfernandez: are these comments a leftover?

carlesfernandez (commented):
Thanks for this contribution! Please check out my inline comments.

src/algorithms/signal_source/adapters/file_source_base.cc
castle055: What exactly are you referring to here? If it's the assignment in line 215, I did it so that `output` correctly reflects which block is the right most block inside the file source. I know it is not used as of now, but if future development adds another block to the file source, this should prevent a headache or two since it makes it explicit that the block to connect it to is the sensor_data_source (if active).

src/algorithms/signal_source/adapters/file_source_base.cc
castle055: Same as in the connect function. The assignment in line 272 is useless right now but could be handy if future development adds more internal blocks to the file source base.

src/algorithms/libs/sensor_data/sensor_identifier.cc
castle055: I thought about it back when I wrote this and I am not convinced. I would need two maps, one from value to string and the other from string to value. That's still two coupled lists in the code. Also, I am trying to avoid hashing those strings if possible. I'm pretty confident that checking equality is faster as long as the number of sensors doesn't grow too large. This function is called every time a sensor sample reaches the `SensorDataAggregator` and I suspect using a map would be slower.

castle055: I also thought about using the [X macro pattern](https://en.wikipedia.org/wiki/X_macro). It would make it easier to maintain as there would only be one list. However, preprocessor magic is often frowned upon and, even though I don't entirely agree with the arguments against it, I try to avoid it when possible.

castle055: I had to look this up because I wasn't sure, but std::unordered_map uses dynamic memory (heap) even if you declare the variable as `static const`. That introduces indirection to an operation that needs to be as fast as possible.

I have not measured the performance impact of such indirection, it may be negligible. However, I'm not a fan of using a dynamically sized container type to store static data.

A static array could be used, but that would downgrade every O(1) function to O(n) as switch cases would need to be replaced by iterations.

Ultimately, all these ideas would work with minimal impact, so I'm open to changing it if appropriate. My personal choice would be using the X macro pattern, but again, not everyone likes using the preprocessor for anything other than feature flags and header guards.

carlesfernandez: ok, let's leave it as is by now.

src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
carlesfernandez: ok, fixed in 8779b13

src/algorithms/libs/sensor_data/sensor_data_type.cc
carlesfernandez: Fixed in 91cdb728f87e8c11beed34f29c9517ccb692d489

src/algorithms/libs/sensor_data/sensor_data_type.cc
carlesfernandez: Fixed in 91cdb728f87e8c11beed34f29c9517ccb692d489

src/algorithms/libs/sensor_data/sensor_data_type.cc
carlesfernandez: Fixed in 91cdb728f87e8c11beed34f29c9517ccb692d489

src/algorithms/signal_source/adapters/file_source_base.cc
carlesfernandez: But this would instantiate another SensorDataSource object, isn't it?

src/algorithms/signal_source/adapters/file_source_base.cc
carlesfernandez: Same as above, I do not understand why are you calling the constructor again. Maybe I'm missing something.

src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc
carlesfernandez: Seems fixed in d97d93ba2181763d791ff0e4940b7f7ac023a950

src/algorithms/signal_source/adapters/file_source_base.cc
carlesfernandez: Fixed in 9baf24392a514e5178b311a2401834039b138c57

src/algorithms/signal_source/adapters/file_source_base.cc
carlesfernandez: Fixed in 9baf24392a514e5178b311a2401834039b138c57

src/algorithms/libs/sensor_data/sensor_data_source_configuration.cc
carlesfernandez: Fixed in 8da4dd1d215beb33ff87845b7b89d625569844ee

src/algorithms/libs/sensor_data/sensor_data_aggregator.cc
carlesfernandez: Fixed in 8da4dd1d215beb33ff87845b7b89d625569844ee

src/algorithms/libs/sensor_data/sensor_data_file.h
carlesfernandez: Fixed in a2c9fa825a46cb6ecb6e9adf6c2606ef05047f2c

src/algorithms/signal_source/adapters/file_source_base.cc
carlesfernandez: Fixed in a2c9fa825a46cb6ecb6e9adf6c2606ef05047f2c

src/algorithms/libs/sensor_data/sensor_data_file.h
carlesfernandez: Fixed in 9fd3d416efc6166be6168be263a9135ea78a51e4

src/algorithms/libs/sensor_data/sensor_data_file.cc
carlesfernandez: Fixed in 9fd3d416efc6166be6168be263a9135ea78a51e4

src/algorithms/signal_source/adapters/file_source_base.cc
castle055: Ah, no. That is not the constructor. That is a getter in FileSourceBase. It's similar to the already present `file_source()`, `valve()`, `throttle()` or `sink()` methods.

.gitignore
```diff
@@ -3,6 +3,7 @@
 *~
 .*.swp
 /build/
+/.build/
```
jwmelto, line 6: Just curious if `.build` is a personal preference? Using a hidden directory for the build folder is unusual.

src/algorithms/PVT/adapters/rtklib_pvt.cc
```diff
@@ -86,6 +86,10 @@ Rtklib_Pvt::Rtklib_Pvt(const ConfigurationInterface* configuration,
     pvt_output_parameters.system_ecef_pos_sd_m = configuration->property(role + ".kf_system_ecef_pos_sd_m", 2.0);
     pvt_output_parameters.system_ecef_vel_sd_ms = configuration->property(role + ".kf_system_ecef_vel_sd_ms", 0.5);
 
+    // PVT Sensor data settings
+    SensorDataSourceConfiguration sensor_data_configuration{configuration};
```
jwmelto, line 90: `SensorDataSource` is a pretty generic name for a very specific sensor (IMU). It might be better to be specific first and let the abstractions reveal themselves, rather than assuming abstractions that don’t exist.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -1993,7 +2006,7 @@ int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_item
         {
             std::vector<gr::tag_t> tags_vec;
             // time tag from obs to pvt is always propagated in channel 0
-            this->get_tags_in_range(tags_vec, 0, this->nitems_read(0), this->nitems_read(0) + noutput_items);
+            this->get_tags_in_range(tags_vec, 0, this->nitems_read(0), this->nitems_read(0) + noutput_items, pmt::mp("timetag"));
```
jwmelto, line 2009: Taken at face value, this is an improvement; the legacy tag handling was non-idiomatic and fragile. However, since you follow this with another tag handling loop, it would be better to combine them and retrieve/iterate over the tags once.

The legacy code did not match a tag key (which I think is bad, if not wrong), but tried to introspect the value type. It is better to document supported tags, extract them by key, and ignore any others. 

This, of course is just my opinion. While I like what you did, I’m not sure you didn’t break something. Hyrum’s law applies.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -2017,6 +2030,10 @@ int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_item
         }
     // ************ end time tags **************
 
+    std::vector<gr::tag_t> sensor_tags;
+    this->get_tags_in_range(sensor_tags, 0, this->nitems_read(0), this->nitems_read(0) + noutput_items, d_sensor_data_aggregator->SENSOR_DATA_TAG);
```
jwmelto, line 2034: Note `get_tags_in_window` has nicer semantics

src/algorithms/PVT/libs/pvt_conf.h
```diff
@@ -103,6 +103,9 @@ class Pvt_Conf
     double measures_ecef_vel_sd_ms = 0.1;
     double system_ecef_pos_sd_m = 0.01;
     double system_ecef_vel_sd_ms = 0.001;
+
+    // Sensor Data parameters
+    bool kf_use_imu_vel = false;
 };
```
jwmelto, line 109: What does `kf` mean? Following the naming style of this structure, `use_imu_vel` or maybe `use_imu_vel_kf` would be more consistent

src/algorithms/PVT/libs/rtklib_solver.cc
```diff
@@ -1569,6 +1577,14 @@ bool Rtklib_Solver::get_PVT(const std::map<int, Gnss_Synchro> &gnss_observables_
                                 {
                                     arma::vec p = {pvt_sol.rr[0], pvt_sol.rr[1], pvt_sol.rr[2]};
                                     arma::vec v = {pvt_sol.rr[3], pvt_sol.rr[4], pvt_sol.rr[5]};
+                                    if (d_conf.kf_use_imu_vel)
```
jwmelto: This block is identical to the above. Did you consider abstracting the common code to a function so the change only had to be made once?

src/algorithms/libs/sensor_data/sensor_data_aggregator.cc
```diff
@@ -0,0 +1,149 @@
...
+ * \file sensor_data_aggregator.cc
+ * \brief  Aggregates sensor samples from gnu radio stream tags into typed lists for easy access
+ * \author Victor Castillo, 2024. victorcastilloaguero(at).gmail.es
+ *
+ * -----------------------------------------------------------------------------
+ *
+ * GNSS-SDR is a Global Navigation Satellite System software-defined receiver.
+ * This file is part of GNSS-SDR.
+ *
+ * Copyright (C) 2010-2021  (see AUTHORS file for a list of contributors)
```
jwmelto, line 11:
```suggestion
 * Copyright (C) 2025  (see AUTHORS file for a list of contributors)
```

src/algorithms/libs/sensor_data/sensor_data_source.cc
jwmelto, line 1: This is a GNU Radio block; it belongs in the blocks directory

src/algorithms/libs/sensor_data/sensor_data_source_configuration.cc
jwmelto, line 1: I think this is an adapter?

src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc
```diff
@@ -657,6 +657,22 @@ void hybrid_observables_gs::set_tag_timestamp_in_sdr_timeframe(const std::vector
         }
 }
 
+void hybrid_observables_gs::propagate_sensor_data()
+{
+    if (d_sensor_data_tags.empty())
```
jwmelto: This is a surprising construct. Why not simply a `while` loop? do/while seems to be frowned upon in general but here it’s just obtuse.

src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc
```diff
@@ -671,9 +687,21 @@ int hybrid_observables_gs::general_work(int noutput_items __attribute__((unused)
         {
             d_Rx_clock_buffer.push_back(in[d_nchannels_in - 1][0].Tracking_sample_counter);
 
-            // time tags
             std::vector<gr::tag_t> tags_vec;
-            this->get_tags_in_range(tags_vec, d_nchannels_in - 1, this->nitems_read(d_nchannels_in - 1), this->nitems_read(d_nchannels_in - 1) + 1);
+            // Propagate sensor data tags
+            get_tags_in_range(tags_vec, d_nchannels_in - 1, this->nitems_read(d_nchannels_in - 1), this->nitems_read(d_nchannels_in - 1) + 1, pmt::mp("sensor_data"));
+            while (!d_sensor_data_tags.empty())
```
jwmelto, line 689: Or just `d_sensor_data_tags.clear()`?

src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc
```diff
@@ -671,9 +687,21 @@ int hybrid_observables_gs::general_work(int noutput_items __attribute__((unused)
...
+                {
+                    d_sensor_data_tags.pop();
+                }
+            for (const auto &tag : tags_vec)
```
jwmelto, line 693: I’m not clear what you wanted to do here. Are these two loops basically doing `d_sensor_data_tags.assign(tags_vec.begin(), tags_vec.end());`?

src/algorithms/signal_source/adapters/file_source_base.cc
```diff
@@ -249,17 +259,24 @@ void FileSourceBase::disconnect(gr::top_block_sptr top_block)
     // DUMP
     if (sink())
         {
-            top_block->disconnect(std::move(output), 0, sink(), 0);
+            top_block->disconnect(output, 0, sink(), 0);
```
jwmelto, line 262: It turns out, all these disconnect methods work too hard. Rather than trying to unwind all the connections, you can just `top_block->disconnect(output)` and the block is removed from the graph.

src/algorithms/signal_source/adapters/file_source_base.cc
```diff
@@ -478,6 +496,7 @@ gnss_shared_ptr<gr::block> FileSourceBase::file_source() const { return file_sou
 gnss_shared_ptr<gr::block> FileSourceBase::valve() const { return valve_; }
 gnss_shared_ptr<gr::block> FileSourceBase::throttle() const { return throttle_; }
 gnss_shared_ptr<gr::block> FileSourceBase::sink() const { return sink_; }
+SensorDataSource::sptr FileSourceBase::sensor_data_source() const { return sensor_data_source_; }
```
jwmelto, line 499: This is very specific for a base class

jwmelto (commented):
This is a pervasive change for a very specialized use case. When I started reading it, I thought you were actually using an IMU, but it appears that you only support file-based playback of recorded files. 

I’m not sure how this time syncs IMU readings with samples; again, with a groomed data file it may not be an issue, but with an actual device it could (the last one I used updated at 400Hz). 

I most don’t like the integration into file source base class, but it’s starting to grow on me, if properly abstracted to not specify your file based IMU reader. I do really like cleaning up the tag handling. Errant tags in source blocks have crashed my program before.

src/algorithms/signal_source/adapters/file_source_base.cc
carlesfernandez: ok, I see. Thanks!

src/algorithms/PVT/adapters/rtklib_pvt.cc
castle055: That's because that's a pretty generic class. It is appropriately called `SensorDataSource` and not `IMUDataSource`. Nothing about this is specific to IMU sensors, any sensor can be used. Barometers, IMUs, Star trackers, you name it, if there is a GNSS algorithm that can use it to increase fix accuracy, we can implement it. There is no need for more abstractions nor is this overly abstracted. Sensor values are just values, once you abstract then behind an identifier (ie: IMU_VEL_X), it becomes trivial to add more sensors and use them, just add more identifiers.

The fact that all currently supported sensors are IMU related doesn't mean it'll always be like that. It just so happens that the algorithms I'm working with need only IMU data. The proof of concept for the Kalman filter uses IMU velocities, and the VTL implementation I plan on working on only needs IMU data as well.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
castle055: GNU Radio matches keys pretty efficiently already. IIRC it does not have to loop through the list of tags like we would. Merging both loops and managing both sets of keys at the same time would result in more logic on our side instead of the already present logic on GNU Radio's side. I'm not sure what would be faster (probably letting GNU Radio match its own keys as I'm doing) but this separates both tags making it clear that they are very different systems.

I familiarized myself with the sample counter (and thus the timetags and their handling across the program) before changing anything.

src/algorithms/PVT/libs/pvt_conf.h
castle055: It stands for 'Kalman filter', this is a configuration value for the Kalman filter. `kf` is a common abbreviation for it, see the `enable_pvt_kf` option.

The prefix `kf_` indicates that this configuration only affects the KF (if it's enabled).

src/algorithms/PVT/libs/rtklib_solver.cc
castle055: No, I didn't consider it, nor do I think that's a good idea. This is just an assignment, like those for `p` and `v` just above. I could follow Robert C Martin's philosophy and separate this into its own three line meaningless method but I believe thinking in terms of locality of behavior is more appropriate here.

What I just realized I can do is lift these assignments out of the `if (d_pvt_kf.is_initialized() == false)` statement. That would get rid of the duplication while maintaining locality.

src/algorithms/libs/sensor_data/sensor_data_source_configuration.cc
castle055: No, it is not. It's a configuration object. It reads values from the `ConfigurationInterface`, stores them in a structured way and validates them with appropriate error messages.

src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc
castle055: You are right, I think I changed the code at some point and I didn't clean up.

src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc
castle055: `std::queue` does not implement that method.

src/algorithms/observables/gnuradio_blocks/hybrid_observables_gs.cc
castle055: `std::queue` does not implement that method.

castle055:
@jwmelto 

> I’m not sure how this time syncs IMU readings with samples; again, with a groomed data file it may not be an issue, but with an actual device it could (the last one I used updated at 400Hz).

If I understand the code correctly, as far as GNSS-SDR is concerned, time does not exist until acquisition finishes and a satellite reports it. Or rather, time is defined not in seconds but in samples. This is, in fact, the problem that the sample counter and timetags solve if enabled. I didn't want to depend on timetags being enabled, so everything is synchronized using sample counts instead.

Please, check the PR text for a configuration example. You get to dictate when the sensor data file starts being read, as well as how fast. Those two configuration values (`sample_offset` and `sample_period`) are enough to tie every sensor sample to the correct RF sample. Once the sensor sample is put inside a stream tag, GNU Radio takes care of it and generally does a great job at propagating the tags alongside the sample they are tied to (the Tracking and Observables blocks are the exception, manual handling is required there).

On top of that, when using the `SensorDataAggregator` every sensor measurement stores an RF sample count. This represent the point in time when the sensor took that specific measurement.

For instance, your 400Hz IMU can be used without issue. If we assume a 30 Msps RF signal, your `sample_period` for your sensor data file would be:

$$\frac{30000000}{400}\frac{\textnormal{[RF samples/s]}}{\textnormal{[sensor samples/s]}} = 75000 \frac{\textnormal{[RF samples]}}{\textnormal{[sensor sample]}}$$

Note that the `SensorDataSource` is placed after the signal source and before any resampling can be done in the signal conditioner, so the `sample_period` must be calculated from the raw RF sample rate.

And yes, currently this can only read measurements from a data file. It cannot interface with any device. If I tried to do that, the code would explode in size. Think of how big GNSS-SDR already is, and it "only" interfaces with one kind of device (RF frontends). Now, imagine if it also had to interface with specific sensors, there would be more code written for that purpose than there is now in the whole project.

For now, if you need to use a sensor in real time you can try to do it through a unix pipe, since they behave almost exactly like files. I have not tested that though.

castle055:
@jwmelto 

> This is a pervasive change for a very specialized use case. When I started reading it, I thought you were actually using an IMU, but it appears that you only support file-based playback of recorded files.

To be explicitly clear, since it appears you didn't read the PR text, these changes are **not** meant only for dealing with IMU measurements. This can handle any sensor that might be needed in the future as well. The concept of sensor fusion applies quite well to GNSS, mainly with IMUs, but there's potential for other sensors as well. I didn't write an overgeneralized IMU interfacing library, I wrote a sensor data handling system and then used it to handle IMU readings. Very different development flow.

The result is that I can get barometer readings, LiDAR images or whatever your algorithm needs into wherever you algorithm is without having to modify the underlying sensor data architecture. Being able to add new sensors in the future was one very important requirement for my design.

.gitignore
castle055: Well, the binaries are already put somewhere else by default, and the IDE runs CMake for me, so there are very few reasons for me to visit or do anything in this directory. I'd rather not see it.

I've been doing it this way for years, but it is a personal choice and I am willing to take this change out of the PR if asked to.

carlesfernandez:
Dear @castle055,

I'm quite sure @jwmelto read the PR description carefully. In fact, he is highlighting a very relevant issue that could improve this PR with minimal effort, simply by making the abstractions clearer and more general.

This contribution is valuable (please check the pending issue regarding compatibility with older versions of GNU Radio, which are still used in some FPGA-based software stacks). That said, it can be made even better, more maintainable and extensible.

Thank you for your efforts so far, and please take all received comments in a constructive spirit. This kind of discussion is a fundamental part of how open-source collaboration work and, in my opinion, one of the most important aspects. Being able to provide thoughtful, informative responses is essential in collaborative development. If parts of the documentation are unclear or potentially misleading, this is a great opportunity to rephrase and improve them to prevent future confusion.

Please also keep in mind that every GNSS-SDR feature that is ready for use needs to be documented on the website, specifically in this case: https://gnss-sdr.org/docs/sp-blocks/signal-source/  I'll be happy to help with that. Still, improving the code and documentation based on feedback from expert early users before a release is incredibly valuable, and I encourage you to take full advantage of it.

castle055:
Thanks for the feedback, @carlesfernandez.

It seemed to me that some parts of the PR description may have been missed, particularly the sections on how the code is structured to allow adding support for other sensors beyond IMUs. That’s also why I think the current class naming is appropriate, since it's meant to be extensible. I believe this is the issue you mention in the first paragraph, but I'm not sure.

That said, I’ve read through all of @jwmelto’s comments and either clarified my intent or made the suggested changes where I agreed. From your comment, though, I’m wondering if I may have missed something. Would you mind clarifying which specific issues you’re referring to? Both the one you mention in the first paragraph and the one about backwards compatibility. I definitely didn’t intend to break compatibility, so if I have, that’s something I’ll fix before merge.

Regarding documentation, I am happy to write it and submit it for review. However, I must admit I'm not familiar with the process of how to contribute documentation to the GNSS-SDR website. Should I make a PR on the website repo?

carlesfernandez:
Hi @castle055,

specifically, I'm asking to change a line in this file: [src/algorithms/libs/sensor_data/sensor_data_source.cc](https://github.com/gnss-sdr/gnss-sdr/pull/894/files/76bf26caa649b2ceb21bf1e3e1a44b525bc94752#diff-ed2ead3ee93579729d37f6475f5c6631a257e6389f3e8a5c0c6042556a383a78)

In line 58, it should be:
```
sensor_data_files_.emplace(id, gnss_make_shared<SensorDataFile>(...)
```

This is required for building GNSS-SDR in some old environments that still run GNU Radio < v3.9.

castle055:
You are right, @carlesfernandez , I missed that.

I've fixed it in commit https://github.com/gnss-sdr/gnss-sdr/pull/894/commits/993d5ff5f41774b6a5c43eb44cb1836e604be058

carlesfernandez:
Thanks! I think now it's ready to be merged. I'll contact you to prepare the corresponding documentation in the website.

castle055:
Sounds great, thanks!

PR #904, opened 2025-07-23
Remove receiver type

MathieuFavreau:
This is a refactor to remove the receiver type integer, which is not scalable and hard to maintain.
Instead I am using the wrapper class `Signal_Enabled_Flags` to check which signals are enabled.

While there is still room for simplifications in the code, I wanted to do the minimal amount of changes, to make it easier to review.
While doing this exercise, I have found some things that looked like errors to me and fixed them, for example the RINEX file used for BEIDOU was the same as GPS, instead of using the file created for BEIDOU.

I do have some questions about some part of the code that I didn't want to change without being sure I understood right, I am going to write comments on these sections, please address them directly.


src/algorithms/PVT/libs/rtcm_printer.cc
```diff
@@ -225,12 +225,9 @@ Rtcm_Printer::~Rtcm_Printer()
...
-    int32_t rtcm_MSM_rate_ms,
-    int32_t rtcm_MT1019_rate_ms,
-    int32_t rtcm_MT1020_rate_ms,
-    int32_t rtcm_MT1045_rate_ms,
+    uint32_t signal_enabled_flags,
     int32_t rtcm_MT1077_rate_ms,
+    int32_t rtcm_MT1087_rate_ms,
     int32_t rtcm_MT1097_rate_ms,
     bool flag_write_RTCM_MSM_output,
     bool flag_write_RTCM_1019_output,
```
MathieuFavreau, line 233: We can only use the flag variable and don't need the rates here, because if you look in `rtklib_pvt_gs.cc` we already do this:
```
if (d_rtcm_MT1019_rate_ms != 0)  // allows deactivating messages by setting rate = 0
    {
        if (current_RX_time_ms % d_rtcm_MT1019_rate_ms == 0)
            {
                flag_write_RTCM_1019_output = true;
            }
    }
```

Where I am confused, is that `rtcm_MSM_rate_ms` is the same as `rtcm_MT1097_rate_ms`, but we allow to change `rtcm_MT1077_rate_ms` and `rtcm_MT1087_rate_ms` in the config, and there is no "flag" to tell the printer it's time to log. I feel we need to add a flag for these two rates, but I wanted to get feedback before doing this change.

MathieuFavreau:
I found some issues with how I changed some stuff, will keep in draft until it's fixed

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -2346,34 +2349,45 @@ int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_item
...
                                         }
-                                    if (d_rtcm_MT1045_rate_ms != 0)
+                                    if (rtcm_MT1045_enabled)
                                         {
                                             if (current_RX_time_ms % d_rtcm_MT1045_rate_ms == 0)
                                                 {
                                                     flag_write_RTCM_1045_output = true;
                                                 }
                                         }
+
```
MathieuFavreau, line 2390: This TODO says that messages 1077, 1087 and 1097 are not used, which no longer seems the case. I wasn't sure exactly how to handle the MSM flag on top of these 3 message flags, so I didn't add them for now

src/algorithms/PVT/libs/rinex_printer.cc
```diff
@@ -296,1263 +359,616 @@ Rinex_Printer::~Rinex_Printer()
...
+                }
+            else
+                {
+                    rinex_header_written = false;
+                }
+
+            d_rinex_header_written = rinex_header_written;
+        }
+
+    if (d_rinex_header_written && flag_write_RINEX_obs_output)  // The header is already written, we can now log the navigation message data
```
MathieuFavreau, line 555: From here sometimes we update the headers without having any ephemeris and sometime we don't, not sure what the logic is here

src/algorithms/PVT/libs/rinex_printer.cc
```diff
@@ -296,1263 +359,616 @@ Rinex_Printer::~Rinex_Printer()
...
-            log_rinex_nav(navFile, new_bds_eph);
-            break;
-        default:
-            break;
+    const Signal_Enabled_Flags flags(signal_enabled_flags);
+
+    if (flags.check_any_enabled(BDS_B1) ||
+        flags.check_any_enabled(BDS_B3))
+        {
+            log_rinex_nav(navBdsFile, new_bds_eph);
```
MathieuFavreau, line 971: Changed from `navFile` (GPS) to `navBdsFile`

src/algorithms/PVT/libs/rtcm_printer.cc
```diff
@@ -240,1270 +242,146 @@ void Rtcm_Printer::Print_Rtcm_Messages(const Rtklib_Solver* pvt_solver,
...
-                                                                }
-                                                        }
-                                                }
-                                        }
-                                    if (gps_eph_iter != pvt_solver->gps_ephemeris_map.cend())
-                                        {
-                                            Print_Rtcm_MSM(7, gps_eph_iter->second, {}, {}, {}, rx_time, gnss_observables_map, enable_rx_clock_correction, 0, 0, false, false);
-                                        }
+                        }
+                    else
```
MathieuFavreau, line 313: For MSM, I tried to keep the same logic as before, but I am a bit confused why it's necessary.

Why can't we simply take the first ephemeris of each map and use that, instead of going in the gnss observables map and then looking into the ephemeris maps and take the first that match? For the previous cases with one constellation, we only looked into the ephemeris maps, so I really don't see the reason to do this.

carlesfernandez (approved):

PR #908, opened 2025-08-07
Cleanup rinex printer

MathieuFavreau:
This PR has for goal to reduce the amount of code duplication in the RINEX printer.
I decided to stop here as this is already a lot of changes, but this is quite straight forward:

- Moved most private functions with no member access to `.cc` file in unnamed namespace
- Found duplicated code and created functions in unnamed namespace
- Added constructor to be able to use the init list and put some variables `const`
- Cleaned up the destructor

There is still a lot of code duplication, but that's a decent first pass


carlesfernandez:
Hi @MathieuFavreau 

Awesome work, as always! There’s still some more clean-up needed in that implementation, but this is a solid step in the right direction. I’ve replaced the use of `std::optional` by raw pointers / `nullptr`, which is less elegant but allows building when using C++14.

Merging.

carlesfernandez (approved):

PR #885, opened 2025-05-07
Fix tracking same PRN multiple times

MathieuFavreau:
It is currently possible to track the same PRN multiple times.
It has more chances of happening when the number of acquisition channels is close to the number of PRN for a given signal. 

This is a bit of a convoluted example in order to reproduce, but if you assign 30 acquisition channels and track GPS L1/CA, it's easy to reproduce:
```
New GPS NAV message received in channel 10: subframe 4 from satellite GPS PRN 11 (Block III) with CN0=54 dB-Hz
New GPS NAV message received in channel 25: subframe 4 from satellite GPS PRN 06 (Block IIF) with CN0=56 dB-Hz
New GPS NAV message received in channel 19: subframe 4 from satellite GPS PRN 09 (Block IIF) with CN0=50 dB-Hz
New GPS NAV message received in channel 24: subframe 4 from satellite GPS PRN 11 (Block III) with CN0=54 dB-Hz
New GPS NAV message received in channel 11: subframe 4 from satellite GPS PRN 04 (Block III) with CN0=47 dB-Hz
New GPS NAV message received in channel 17: subframe 4 from satellite GPS PRN 19 (Block IIR) with CN0=51 dB-Hz
New GPS NAV message received in channel 2: subframe 4 from satellite GPS PRN 29 (Block IIR-M) with CN0=46 dB-Hz
New GPS NAV message received in channel 20: subframe 4 from satellite GPS PRN 12 (Block IIR-M) with CN0=55 dB-Hz
New GPS NAV message received in channel 9: subframe 4 from satellite GPS PRN 04 (Block III) with CN0=47 dB-Hz
New GPS NAV message received in channel 13: subframe 4 from satellite GPS PRN 05 (Block IIR-M) with CN0=51 dB-Hz
New GPS NAV message received in channel 8: subframe 4 from satellite GPS PRN 09 (Block IIF) with CN0=50 dB-Hz
New GPS NAV message received in channel 15: subframe 4 from satellite GPS PRN 05 (Block IIR-M) with CN0=51 dB-Hz
New GPS NAV message received in channel 0: subframe 4 from satellite GPS PRN 25 (Block IIF) with CN0=48 dB-Hz
New GPS NAV message received in channel 7: subframe 4 from satellite GPS PRN 25 (Block IIF) with CN0=48 dB-Hz
New GPS NAV message received in channel 1: subframe 4 from satellite GPS PRN 06 (Block IIF) with CN0=56 dB-Hz
```

[...]

PR #910, opened 2025-08-12
Fix: Prevent PRN reassignment before Acquisition

pedromiguelcp:
This pull request fixes two related issues:

- **PRNs assigned at initialization are not acquired** — introduced in [`72a4ed3`](https://github.com/gnss-sdr/gnss-sdr/commit/72a4ed3878227a69c3090fbaec99211887fbbb93)  
- **Same PRN assigned to multiple channels** — described in [PR #885](https://github.com/gnss-sdr/gnss-sdr/pull/885)

---

## Context

Before commit [`72a4ed3`](https://github.com/gnss-sdr/gnss-sdr/commit/72a4ed3878227a69c3090fbaec99211887fbbb93), the same PRN could be assigned to multiple channels.  
This required a corner-case configuration (high number of acquisition blocks) but was possible.

That commit solved the duplicate-assignment problem by changing the logic to **not push back** PRNs assigned to channels into the available list.  
While this fixed the multi-assignment issue, it **introduced a new problem**:

> When assigning PRNs to channels during initialization, those PRNs can be “forgotten” if, before acquisition is executed and reported for that channel, another PRN is assigned to the same channel.

This happens because `search_next_signal()` removes PRNs from the available list as soon as they’re assigned.  
In configurations with one or few acquisition blocks, the initial PRNs assigned to channels may never be acquired.

---

## Proposed Solution

Introduce a new **Assigned** channel state for PRNs that have been allocated to a channel but have not yet started acquisition.

A new PRN will only be assigned to a channel when an acquisition result (success or fail) has been reported for its current PRN.

  - **Previous states:**  
    `0` – Idle  
    `1` – Acquisition  
    `2` – Tracking  
  - **New states:**  
    `0` – Idle  
    `1` – Assigned (PRN allocated but acquisition not yet started)  
    `2` – Acquisition  
    `3` – Tracking

### DIFF
--- a/src/core/receiver/gnss_flowgraph.cc
+++ b/src/core/receiver/gnss_flowgraph.cc
@@ -571,7 +571,7 @@ int GNSSFlowgraph::connect_desktop_flowgraph()
     for (int i = 0; i < channels_count_; i++)
         {
             LOG(INFO) << "Channel " << i << " assigned to " << channels_.at(i)->get_signal();
-            if (channels_state_[i] == 1)
+            if (channels_state_[i] == 2)
                 {
 #if ENABLE_FPGA
                     if (enable_fpga_offloading_)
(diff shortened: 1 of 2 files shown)


carlesfernandez:
Good catch! Thanks @pedromiguelcp

PR #911, opened 2025-08-13
Skyplot utility: fixes and new `--use-obs` feature

pedromiguelcp:
This Pull Request suggests two fixes and a feature for the skyplot util.

## Fixes
1. **Duplicate ephemeris time subtraction**  
   In `calculate_satellite_positions()`, the ephemeris reference time was being subtracted from `transmit_time` which already had the ephemeris reference time subtracted, causing large timing offsets, incorrect az/el values, and exclusion of visible/tracked satellites.  
   → Removed the duplicated subtraction.

2. **Ephemeris selection**  
   The latest ephemeris was always used, which is accurate for epochs near that reference time but introduces errors for distant epochs.  
   → Now selects the ephemeris whose reference time is closest to the midpoint of the receiver observation time window.

## Feature
- **`--use-obs` optional flag**  
  Allows plotting only for the receiver observation period by using a matching RINEX observation file. Falls back to NAV (ephemeris) time range if no matching OBS is found.

## Changes
- **Fix:** Removed duplicate subtraction of ephemeris reference time.  
- **Fix:** Select ephemeris closest to observation midpoint for better accuracy.  
- **Feat:** Added `--use-obs` option to limit plot to receiver observation period.

### DIFF
--- a/utils/skyplot/README.md
+++ b/utils/skyplot/README.md
@@ -16,8 +16,10 @@ showing satellite visibility over time.
 ## Features
 
 - Processes RINEX navigation files.
-- Optionally uses an OBS file to limit plot to the receiver active pricessing time (--use-obs).
-  - When enabled, the tool looks for a matching file by replacing the last character of the NAV filename with O/o and uses it if found.
+- Optionally uses an OBS file to limit plot to the receiver active processing 
+  time (--use-obs).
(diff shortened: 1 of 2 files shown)


carlesfernandez:
Thanks!

PR #916, opened 2025-08-31
Fixes and Improvements for Sensor Data Handling

castle055:
This PR is part of my [Google Summer of Code 2025 project](https://gist.github.com/castle055/902862007ac1d97f35702bc51347b549).

After spending the summer using the sensor data infrastructure (merged in https://github.com/gnss-sdr/gnss-sdr/pull/894), I have found a few issues with the implementation. This PR solves these issues and also adds a couple minor features.

## Changes

- **ADDED:** `SensorDataType::F64` implements double floating point precision. IMU measurements now default to this data type instead of `F32`.
- **ADDED:** `SensorDataAggregator::get_average_<<type>>(SensorIdentifier)` returns the average among all measurements in the current update cycle.
- **FIX:** Off-by-one error within the `SensorDataFile`. Instead of storing the last RF sample with a measurement, the **next** RF sample to be tagged is stored, shifting all measurement by one period into the past.
- **FIX:** Resampler blocks now update the timestamp stored in each sensor measurement to reflect the interpolation/decimation performed on the RF stream.
- **FIX:** The `HybridObservables` block now delays sensor samples in order to match the delay caused by tracking and telemetry decoding. The delay is not fixed, measurements are pushed into a queue as they arrive from the sample counter block and only released when a channel outputs a `Gnss_Synchro::Tracking_sample_counter` larger than the one stored within the sensor measurement. This means the delay adapts in real time to the computing power available maintaining synchronicity.
- **FIX:** Sometimes the math in `SensorDataSource` added up and sensor measurements where tagged into RF samples that were already gone from the block's buffer. I misunderstood the way GNU Radio handled tags and thought this was okay. In reality, a block can only tag samples while they are in its buffers and so those sensor measurements were being lost. The solution is to tag the measurements to the first RF sample in the buffer, whichever that may be. This should not introduce any timing errors as sensor measurements and `Gnss_Synchro` objects are matched by the timestamp stored inside the sensor measurement, not by the offset of the RF sample being tagged.

### DIFF
--- a/src/algorithms/libs/sensor_data/sensor_data_resampler.cc
+++ b/src/algorithms/libs/sensor_data/sensor_data_resampler.cc
@@ -1,7 +1,7 @@
 /*!
  * \file sensor_data_resampler.cc
  * \brief  Updates timestamp within sensor data tags. To be used in resampler blocks.
- * \author Victor Castillo, 2024. victorcastilloaguero(at)gmail.com
+ * \author Victor Castillo, 2025. victorcastilloaguero(at)gmail.com
  *
  * -----------------------------------------------------------------------------
  *
--- a/src/algorithms/libs/sensor_data/sensor_data_resampler.cc
+++ b/src/algorithms/libs/sensor_data/sensor_data_resampler.cc
@@ -25,7 +25,9 @@ std::vector<gr::tag_t> resample_sensor_data_tags(const std::vector<gr::tag_t>& t
         {
             if (pmt::dict_has_key(tag.value, SAMPLE_STAMP_KEY))
                 {
-                    auto& new_tag = new_tags.emplace_back();
+                    new_tags.emplace_back();
+                    auto& new_tag = new_tags.back();
+
                     uint64_t sample_stamp = pmt::to_uint64(pmt::dict_ref(tag.value, SAMPLE_STAMP_KEY, pmt::from_uint64(0)));
                     sample_stamp = sample_stamp * freq_out / freq_in;
                     new_tag.offset = tag.offset * freq_out / freq_in;
--- a/src/algorithms/libs/sensor_data/sensor_data_resampler.cc
+++ b/src/algorithms/libs/sensor_data/sensor_data_resampler.cc
@@ -1,0 +1,37 @@
+/*!
+ * \file sensor_data_resampler.cc
+ * \brief  Updates timestamp within sensor data tags. To be used within resampler blocks.
+ * \author Victor Castillo, 2024. victorcastilloaguero(at)gmail.com
+ *
+ * -----------------------------------------------------------------------------
+ *
+ * GNSS-SDR is a Global Navigation Satellite System software-defined receiver.
+ * This file is part of GNSS-SDR.
+ *
+ * Copyright (C) 2025  (see AUTHORS file for a list of contributors)
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * -----------------------------------------------------------------------------
+ */
+
+#include "sensor_data_resampler.h"
+#include "sensor_identifier.h"
+
+std::vector<gr::tag_t> resample_sensor_data_tags(const std::vector<gr::tag_t>& tags, double freq_in, double freq_out)
+{
+    static pmt::pmt_t SAMPLE_STAMP_KEY = pmt::mp(SensorIdentifier::to_string(SensorIdentifier::SAMPLE_STAMP));
+    std::vector<gr::tag_t> new_tags{};
+    for (auto& tag : tags)
+        {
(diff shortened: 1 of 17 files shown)


src/algorithms/resampler/gnuradio_blocks/direct_resampler_conditioner_cb.cc
```diff
@@ -18,6 +18,8 @@
 
 
 #include "direct_resampler_conditioner_cb.h"
+#include "../../libs/sensor_data/sensor_data_resampler.h"
```
carlesfernandez: Please replace these two lines by

```
#include "sensor_data/sensor_data_resampler.h"
#include "sensor_data/sensor_identifier.h"
```

and then in `src/algorithms/resampler/gnuradio_blocks/CMakeLists.txt` add the dependency:

```
target_link_libraries(resampler_gr_blocks
    PUBLIC
        Gnuradio::runtime
        Boost::headers   # Fix for homebrew
    PRIVATE
        algorithms_libs
        Volk::volk
)
```

src/algorithms/resampler/gnuradio_blocks/direct_resampler_conditioner_cc.cc
```diff
@@ -18,6 +18,8 @@
 
 
 #include "direct_resampler_conditioner_cc.h"
+#include "../../libs/sensor_data/sensor_data_resampler.h"
```
carlesfernandez: Please replace these two lines by

```
#include "sensor_data/sensor_data_resampler.h"
#include "sensor_data/sensor_identifier.h"
```

and then in `src/algorithms/resampler/gnuradio_blocks/CMakeLists.txt` add the dependency:

```
target_link_libraries(resampler_gr_blocks
    PUBLIC
        Gnuradio::runtime
        Boost::headers   # Fix for homebrew
    PRIVATE
        algorithms_libs
        Volk::volk
)
```

src/algorithms/resampler/gnuradio_blocks/direct_resampler_conditioner_cs.cc
```diff
@@ -18,6 +18,8 @@
 
 
 #include "direct_resampler_conditioner_cs.h"
+#include "../../libs/sensor_data/sensor_data_resampler.h"
```
carlesfernandez: Please replace these two lines by

```
#include "sensor_data/sensor_data_resampler.h"
#include "sensor_data/sensor_identifier.h"
```

and then in `src/algorithms/resampler/gnuradio_blocks/CMakeLists.txt` add the dependency:

```
target_link_libraries(resampler_gr_blocks
    PUBLIC
        Gnuradio::runtime
        Boost::headers   # Fix for homebrew
    PRIVATE
        algorithms_libs
        Volk::volk
)
```

carlesfernandez (commented):
See inline comments

src/algorithms/libs/sensor_data/sensor_data_resampler.cc
```diff
@@ -0,0 +1,37 @@
...
+
+std::vector<gr::tag_t> resample_sensor_data_tags(const std::vector<gr::tag_t>& tags, double freq_in, double freq_out)
+{
+    static pmt::pmt_t SAMPLE_STAMP_KEY = pmt::mp(SensorIdentifier::to_string(SensorIdentifier::SAMPLE_STAMP));
+    std::vector<gr::tag_t> new_tags{};
+    for (const auto& tag : tags)
+        {
+            if (pmt::dict_has_key(tag.value, SAMPLE_STAMP_KEY))
+                {
+                    auto& new_tag = new_tags.emplace_back();
```
carlesfernandez: Please replace this line (only works from C++17) by these lines:

```
new_tags.emplace_back();
auto& new_tag = new_tags.back();
```

to ensure C++14 compatibility.

castle055:
Thanks @carlesfernandez for the feedback. I've applied all suggested changes.

src/algorithms/libs/sensor_data/sensor_data_source.cc
```diff
@@ -115,7 +115,7 @@ int SensorDataSource::work(int noutput_items,
                             pmt::pmt_t value = SensorIdentifier::convert_to_internal_type(sensor.identifier, sensor.type, raw_value);
                             data_tag = pmt::dict_add(data_tag, sensor.tag_key, value);
                         }
-                    add_item_tag(0, sample_stamp, TAG_KEY, data_tag);
+                    add_item_tag(0, nitems_written(0), TAG_KEY, data_tag);
```
carlesfernandez, line 116: In this file, line:

```
using namespace std::string_literals;
```

is not used/needed. Please remove it.

Issue #922, opened 2025-09-22
Labsat 4


MathieuFavreau:
Any plans to support Labsat 4 files?

PR #925, opened 2025-09-29
Initial Labsat 4 support

MathieuFavreau:
Closes issue #922, added support for Labsat 4 files (.LS4).

I only tested with 2 bit quantization and 2 channels, but it should work with 1, 2, 3, 4, 8 bit quantization and 1 to 3 channels.
Will probably need another PR to update the code to support 12 bit quantization.
Need to figure out the binary layout with 12 bits which isn't a divisor of 64.

I've added the config parameter `.seconds_to_skip` only for LS4 files.


carlesfernandez:
Applied minor changes to make it buildable with C++11, and added the documentation to the website. Thanks!

PR #932, opened 2025-10-21
Handle channels with different bandwidths for Labsat 4 files

MathieuFavreau:
- Created a mapping of channel index to channel state to generalize code
- Implemented computation of file duration and seconds to skip with LS4 files when all channels do not have the same bandwidth

Note: It can now read files with channels using different bandwidths, but selecting multiple channels with different bandwidths is not supported and will result in an error message


carlesfernandez:
Added my usual, somewhat romantic “C++11 fix” (I don’t expect it to have any impact on performance at all) and included you as an author in the file header. Totally deserved! 😄

PR #933, opened 2025-10-22
Cleanup pcps acquisition

MathieuFavreau:
I have been debugging the `.blocking=false` parameter for the acquisition block and noticed it doesn't seem to work as intended.
Looking at [this commit](https://github.com/gnss-sdr/gnss-sdr/commit/3bcac95964ecb3c53e863124bf9874b6e30ebe18), it seems  while trying to fix a race condition the `.blocking=false` feature was broken.

When `.blocking=false` is set, the acquisition is done on another thread, but since that commit the thread keeps the lock for the whole process of the acquisition, which basically makes it synchronous again.

While trying to understand the access of the member variables between the different threads to see where and how I can release the lock, I've made some changes to the code to make it a little bit easier to understand:

- Removed some duplicated code
- Added `const` to member variables that didn't change (no data race on `const` variables)
- Removed some member variables that didn't need to be

I think optimally a bunch of other member variables could be set at construction and set `const`, but this requires changes to other parts of the codebase, therefor I will keep this for another MR.

### DIFF
--- a/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
+++ b/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
@@ -65,8 +65,8 @@ pcps_acquisition::pcps_acquisition(const Acq_Conf& conf_)
       d_threshold(0.0),
       d_mag(0),
       d_input_power(0.0),
-      d_test_statistics(0.0),
       d_doppler_center_step_two(0.0),
+      d_doppler_max(conf_.doppler_max),
       d_state(0),
       d_positive_acq(0),
       d_doppler_center(0U),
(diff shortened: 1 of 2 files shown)


carlesfernandez:
Very nice refactoring! And thanks for raising the issue with the `.blocking=false` parameter, I think you’ve spotted the very source of the problem. Apparently, I messed it up while trying to fix another issue. I’m running some tests to ensure everything behaves as before, and if all goes well, I’ll merge this.

MathieuFavreau:
> Very nice refactoring! And thanks for raising the issue with the `.blocking=false` parameter, I think you’ve spotted the very source of the problem. Apparently, I messed it up while trying to fix another issue. I’m running some tests to ensure everything behaves as before, and if all goes well, I’ll merge this.

Perfect, just to be clear I did not address the blocking issue in that PR if that wasn't clear

PR #935, opened 2025-10-24
Cleanup RINEX obs headers

MathieuFavreau:
Should give the exact same behavior except:
- I changed the constellation legend in `rinex_obs_header` for beidou because it was more then 60 characters
- I removed one of the two marker type in `rinex_obs_header` for `MIXED (GPS/GLO)`


carlesfernandez:
Thanks for your work on this. I appreciate the comments about allowing customization of headers, that’s definitely something to consider implementing through configuration parameters and/or flags.

PR #938, opened 2025-11-04
More cleanup acquisition and tracking classes

MathieuFavreau:
Let's keep going with cleaning up acquisition and tracking classes, @carlesfernandez great to see you clean it up as well!

I have a question regarding code like this:
```
if (item_type_ == "gr_complex")
    {
        ...
    }
else
    {
        item_size_ = 0;
        acquisition_cc_ = nullptr;
        LOG(WARNING) << item_type_ << " unknown acquisition item type";
    }
```

What are we expecting when `item_type_ != "gr_complex"`?
To me this is just an invalid configuration and we should throw an error to the user, not ignore it, but interested to see other opinions.

### DIFF
--- a/src/algorithms/acquisition/adapters/galileo_e1_pcps_ambiguous_acquisition.cc
+++ b/src/algorithms/acquisition/adapters/galileo_e1_pcps_ambiguous_acquisition.cc
@@ -26,6 +26,7 @@ GalileoE1PcpsAmbiguousAcquisition::GalileoE1PcpsAmbiguousAcquisition(
     unsigned int in_streams,
     unsigned int out_streams)
     : BasePcpsAcquisition(configuration, role, in_streams, out_streams, GALILEO_E1_CODE_CHIP_RATE_CPS, GALILEO_E1_OPT_ACQ_FS_SPS, GALILEO_E1_B_CODE_LENGTH_CHIPS, 4),
+      configuration_(configuration),
       acquire_pilot_(configuration->property(role + ".acquire_pilot", false))
 {
 }
--- a/src/algorithms/acquisition/adapters/galileo_e1_pcps_ambiguous_acquisition.cc
+++ b/src/algorithms/acquisition/adapters/galileo_e1_pcps_ambiguous_acquisition.cc
@@ -32,6 +32,13 @@ GalileoE1PcpsAmbiguousAcquisition::GalileoE1PcpsAmbiguousAcquisition(
 }
 
 
+void GalileoE1PcpsAmbiguousAcquisition::set_gnss_synchro(Gnss_Synchro* p_gnss_synchro)
+{
+    gnss_synchro_ = p_gnss_synchro;
+    BasePcpsAcquisition::set_gnss_synchro(p_gnss_synchro);
+}
+
+
 void GalileoE1PcpsAmbiguousAcquisition::set_channel(unsigned int channel)
 {
     cboc = configuration_->property("Acquisition" + std::to_string(channel) + ".cboc", false);
@@ -49,7 +56,10 @@ void GalileoE1PcpsAmbiguousAcquisition::code_gen_complex_sampled(own::span<std::
         }
     else
         {
-            const std::array<char, 3> signal_str = {{'1', 'B', '\0'}};
-            galileo_e1_code_gen_complex_sampled(dest, signal_str, cboc, prn, sampling_freq, 0, false);
+            std::array<char, 3> Signal_{};
+            Signal_[0] = gnss_synchro_->Signal[0];
+            Signal_[1] = gnss_synchro_->Signal[1];
+            Signal_[2] = '\0';
+            galileo_e1_code_gen_complex_sampled(dest, Signal_, cboc, prn, sampling_freq, 0, false);
         }
 }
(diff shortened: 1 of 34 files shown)


src/algorithms/acquisition/adapters/galileo_e1_pcps_ambiguous_acquisition.cc
```diff
@@ -17,272 +17,38 @@
...
-                {
-                    galileo_e1_code_gen_complex_sampled(code, Signal_,
-                        cboc, gnss_synchro_->PRN, acq_parameters_.resampled_fs, 0, false);
-                }
-            else
-                {
-                    galileo_e1_code_gen_complex_sampled(code, Signal_,
-                        cboc, gnss_synchro_->PRN, fs_in_, 0, false);
-                }
+            const std::array<char, 3> signal_str = {{'1', 'B', '\0'}};
```
MathieuFavreau: Should always be `1B`

src/algorithms/acquisition/adapters/galileo_e1_pcps_ambiguous_acquisition.cc
```diff
@@ -17,272 +17,39 @@
...
-
-    acquisition_->set_gnss_synchro(gnss_synchro_);
 }
 
 
-signed int GalileoE1PcpsAmbiguousAcquisition::mag()
+void GalileoE1PcpsAmbiguousAcquisition::set_channel(unsigned int channel)
 {
-    return acquisition_->mag();
+    cboc = configuration_->property("Acquisition" + std::to_string(channel) + ".cboc", false);
```
MathieuFavreau: Isn't this a bit odd, shouldn't it be `role + std::to_string(channel) + ".cboc"` perhaps?

jwmelto:
@MathieuFavreau  I think the code you questioned is a pattern that is commonly used in the adapter classes. Some adapters support different data types, so the if/(else if)/else block makes more sense there. The flowgraph will go through validation after the blocks have been created and the failure to instantiate the block will be detected and the flowgraph terminate (at least, I think that's the theory).

There are different theories for error handling. I think this code base follows the idea of "detect all the errors, report, then terminate". I suspect I had the same questions as you when I first started working on it.

MathieuFavreau:
> @MathieuFavreau I think the code you questioned is a pattern that is commonly used in the adapter classes. Some adapters support different data types, so the if/(else if)/else block makes more sense there. The flowgraph will go through validation after the blocks have been created and the failure to instantiate the block will be detected and the flowgraph terminate (at least, I think that's the theory).
> 
> There are different theories for error handling. I think this code base follows the idea of "detect all the errors, report, then terminate". I suspect I had the same questions as you when I first started working on it.

I was just thinking, if the outcome is the same (report the error to user), at least reporting it as soon as possible makes it that we don't have to check in all function for `if (item_type_ == "gr_complex")` by fear it is called before we abort, and would simplify the code.

carlesfernandez:
Well, that was probably not the most elegant solution to handle configurations that cannot be instantiated. We opted not to throw in the constructors, which is not inherently bad, but would imply a redesign of how these kinds of “impossible configurations” are handled at the receiver level. Instead, we set the `item_size` to 0 and caught that in the block factory (see [this line](https://github.com/gnss-sdr/gnss-sdr/blob/6d45520f2521b11ca056bd7baee515c382403edb/src/core/receiver/gnss_block_factory.cc#L899)
). It’s admittedly an ugly solution, but it works.

MathieuFavreau:
> Well, that was probably not the most elegant solution to handle configurations that cannot be instantiated. We opted not to throw in the constructors, which is not inherently bad, but would imply a redesign of how these kinds of “impossible configurations” are handled at the receiver level. Instead, we set the `item_size` to 0 and caught that in the block factory (see [this line](https://github.com/gnss-sdr/gnss-sdr/blob/6d45520f2521b11ca056bd7baee515c382403edb/src/core/receiver/gnss_block_factory.cc#L899) ). It’s admittedly an ugly solution, but it works.

Thanks for sharing where in the code this is handled! I mainly wanted to understand the current logic and know if you had considered the solution I mentioned previously and if there were reasons it could not work. I totally understand decisions were made in the past and that there is not always time to refactor codebases, you satisfied my curiosity thanks!

src/algorithms/acquisition/adapters/galileo_e1_pcps_ambiguous_acquisition.cc
carlesfernandez: Oh, the cboc feature was disconnected for ages and no one noticed. I think we need `role() + ".cboc"` there. The refactoring helps to spot these issues 😊

MathieuFavreau: Even better, I did wonder why it used the channel number, was that an older way of setting parameters in the config?
I changed it and took the time to apply the same fix to all the other E1 acquisition classes.

MathieuFavreau:
Just a heads up, I have a bunch of other branches following this PR, I try to cut the work in smaller pieces to make it easier to review as I go

src/algorithms/acquisition/adapters/galileo_e1_pcps_ambiguous_acquisition.cc
carlesfernandez: Thanks! Indeed, that was the older approach to setting parameters per channel.

PR #940, opened 2025-11-06
Make doppler max const and remove setter

MathieuFavreau:
In my opinion it is easier to understand code and less error prone when the state is set at object construction in const member variables, if they aren't going to change during the object lifetime.

This is especially true for multi-threaded code, since we can assume const member variables won't have race conditions (if we make the sane assumption that no-one is using const cast).


MathieuFavreau:
@carlesfernandez I have made a base class for the other acquisition classes in another branch, just so you know to limit duplicated efforts

PR #942, opened 2025-11-10
Add other acquisition base class temporarily to remove duplicated code

MathieuFavreau:
For now I've added a new class `BasePcpsAcquisitionCustom` (yeah I know, didn't have much inspiration) which we should be able to merge with `BasePcpsAcquisition`, but for now kept separated as a first step.

I've added an interface `acquisition_impl_interface` in order for `BasePcpsAcquisitionCustom` to be able to generalize access to the acquisition implementation classes.

I am not sure about `GpsL1CaPcpsAcquisitionFineDoppler`, the previous implementation was reading `.coherent_integration_time_ms` from the config but never used it, can the implementation work with integration times larger then 1ms? The documentation list that parameter so that's why I am unsure.

Also I wanted to use this base class with `GalileoE1PcpsQuickSyncAmbiguousAcquisition` and `GpsL1CaPcpsQuickSyncAcquisition`, but I am wondering if there might be bugs in the implementations and wanted to ask before I touch them:
- In `GpsL1CaPcpsQuickSyncAcquisition`, `code_` is initialized a size `code_length_` but the span in `set_local_code` is initialized at size `vector_length_`, which is `code_length_ * sampled_ms_`. With the test `GpsL1CaPcpsQuickSyncAcquisitionGSoC2014Test` it isn't an issue since `sampled_ms_` and `folding_factor_` are equal, so `for (unsigned int i = 0; i < (sampled_ms_ / folding_factor_); i++)` will only access the allocated code. Should `code_` be initialized with size `vector_length_`, and `vector_length_` actually be `code_length_ * sampled_ms_ / folding_factor_`?
- Same issue for `GalileoE1PcpsQuickSyncAmbiguousAcquisition`

### DIFF
--- a/src/algorithms/acquisition/adapters/CMakeLists.txt
+++ b/src/algorithms/acquisition/adapters/CMakeLists.txt
@@ -7,6 +7,7 @@
 
 set(ACQ_ADAPTER_SOURCES
     base_pcps_acquisition.cc
+    base_pcps_acquisition_custom.cc
     gps_l1_ca_pcps_acquisition.cc
     gps_l1_ca_pcps_assisted_acquisition.cc
     gps_l1_ca_pcps_acquisition_fine_doppler.cc
@@ -31,6 +32,7 @@ set(ACQ_ADAPTER_SOURCES
 
 set(ACQ_ADAPTER_HEADERS
     base_pcps_acquisition.h
+    base_pcps_acquisition_custom.h
     gps_l1_ca_pcps_acquisition.h
     gps_l1_ca_pcps_assisted_acquisition.h
     gps_l1_ca_pcps_acquisition_fine_doppler.h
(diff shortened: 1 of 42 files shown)


MathieuFavreau:
@carlesfernandez for the QuickSync implementations I think I have the fix, we can look at it in another PR

carlesfernandez:
I like the idea, thanks for your work!

Regarding `GpsL1CaPcpsAcquisitionFineDoppler`, I’m afraid it can only operate with `coherent_integration_time_ms = 1`. I’m missing a way to override this parameter in the constructor, since the current implementation will fail if the user sets coherent_integration_time_ms > 1 (it was already failing before, so no regression here).

For the QuickSyncAcquisition, I think you’re right: `vector_length_` should indeed be `code_length_ * sampled_ms_ / folding_factor_`.

PR #943, opened 2025-11-11
Cleanup and fix QuickSync acquisition

MathieuFavreau:
As discussed in this [PR](https://github.com/gnss-sdr/gnss-sdr/pull/942), there was an issue with the QuickSync classes where the size of the code wasn't correct. As we noted in that discussion, `vector_length_` should be `code_length_ * sampled_ms_ / folding_factor_`, or actually `code_length_ * sampled_ms_ / (folding_factor_ * ms_per_code)` for the more generic case (`code_period_ms` is 1 for GPS L1CA).

Using the `BasePcpsAcquisitionCustom` base class constructor, when setting `ms_per_code` I multiply it by the `folding_factor`, so that `ms_per_code` is actually `ms_per_code * folding_factor`. It will then force `sampled_ms` to a multiple of `ms_per_code` just like it used to, and set the `vector_length` to what we defined above.

It seems the size passed in `pcps_quicksync_acquisition_cc` constructor was off to match the incorrect buffer size. I simply passed the size of the vector which we know to be the correct size to fix that issue.

### DIFF
--- a/src/algorithms/acquisition/adapters/gps_l1_ca_pcps_quicksync_acquisition.cc
+++ b/src/algorithms/acquisition/adapters/gps_l1_ca_pcps_quicksync_acquisition.cc
@@ -64,8 +64,8 @@ GpsL1CaPcpsQuickSyncAcquisition::GpsL1CaPcpsQuickSyncAcquisition(
             const unsigned int max_dwells = acq_parameters_.bit_transition_flag ? 2 : acq_parameters_.max_dwells;
 
             acquisition_cc_ = pcps_quicksync_make_acquisition_cc(folding_factor_,
-                vector_length_, max_dwells, acq_parameters_.doppler_max, acq_parameters_.doppler_step, acq_parameters_.fs_in, code_length_, acq_parameters_.bit_transition_flag,
-                acq_parameters_.dump, acq_parameters_.dump_filename, acq_parameters_.enable_monitor_output);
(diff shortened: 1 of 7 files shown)


src/algorithms/acquisition/adapters/gps_l1_ca_pcps_quicksync_acquisition.cc
```diff
@@ -19,250 +19,70 @@
...
+namespace
+{
+uint32_t get_folding_factor(const ConfigurationInterface* configuration, const std::string& role)
+{
+    const int64_t fs_in_deprecated = configuration->property("GNSS-SDR.internal_fs_hz", static_cast<int64_t>(4000000));
+    const auto fs_in = configuration->property("GNSS-SDR.internal_fs_sps", fs_in_deprecated);
+    const auto code_length = static_cast<unsigned int>(round(fs_in / (GPS_L1_CA_CODE_RATE_CPS / GPS_L1_CA_CODE_LENGTH_CHIPS)));
+    const auto folding_factor = static_cast<unsigned int>(ceil(sqrt(log2(code_length))));
+    return configuration->property(role + ".folding_factor", folding_factor);
+}
```
MathieuFavreau, line 41: Having to call this twice and read the config multiple time for the same properties isn't ideal, but I think is good enough for now

PR #944, opened 2025-11-12
Last acquisition classes cleanup

MathieuFavreau:
These are the two last acquisition classes which were not using the new base classes


carlesfernandez:
That was quite a journey!
Thank you for the excellent work and for making the review process so smooth.

PR #948, opened 2025-11-18
Remove init from acquisition

MathieuFavreau:
The `init` function is called in the `Channel` constructor right after the construction of the acquisition classes, and now that the doppler member variables are set at the acquisition construction and are `const`, there is no longer the risk that they get changed before `init` is call, which means we can move the work done in `init` to the acquisition classes constructors.


MathieuFavreau:
@carlesfernandez after this MR I will open another one to remove `set_threshold`, and then I should be good to fix the `blocking=false` issue with `pcps_acquisition` we previously discussed

PR #951, opened 2025-11-24
Cleanup pcps acquisition

MathieuFavreau:
Final cleanup of pcps acquisition class before I fix the blocking issue:
- Make more members `const`
- Sort members by `const` and `non-const`
- Remove code duplication
- Helper functions to write matlab dump


carlesfernandez:
I really like the `matlab_writter_helper.h` ❤️

PR #952, opened 2025-11-25
Fix acquisition blocking param

MathieuFavreau:
Ok so this isn't as clean as I would have hoped for, but it fixes the bug where the `blocking` parameter is blocking even when set to `false`, and it should be thread-safe.

The biggest change is moving the worker thread to a member variable, and making sure to join it in the destructor.
This should fix potential problems where the `pcps_acquisition` destructor is cleaned up while the worker thread is accessing the member variables, which could lead to a crash.

I also make sure to join the worker thread when `pcps_acquisition` is disactivated, because while it's inactive the flowgraph can make changes to the class which could also lead to race conditions.


carlesfernandez:
Epic bug fix! 😊

PR #955, opened 2025-12-01
Fix acquisition two steps

MathieuFavreau:
Well I went a little bit too fast and didn't realize there wasn't any tests for `make_two_steps=true`.
The input power needs to stay in a member variable if we don't want to re-compute it on the second step, which I had not done.

I wanted to add a test, but I think we would need a test signal longer then 2ms to get an acquisition with `make_two_steps`.

### DIFF
--- a/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
+++ b/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
@@ -132,6 +132,7 @@ pcps_acquisition::pcps_acquisition(const Acq_Conf& conf_)
       d_worker_active(false),
       d_num_noncoherent_integrations_counter(0),
       d_dump_number(0),
+      d_input_power(0),
       d_doppler_center_step_two(0),
       d_magnitude_grid(d_num_doppler_bins, volk_gnsssdr::vector<float>(d_fft_size)),
       d_tmp_buffer(d_fft_size),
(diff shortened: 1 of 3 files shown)


MathieuFavreau:
@carlesfernandez I am not aware of the process to add `signal_samples` for the tests, if you can help with that (if you think we should add a test for `make_two_steps=true`)

carlesfernandez:
Hi @MathieuFavreau,

Here you have an example of how to add a new input file for testing:
https://github.com/gnss-sdr/gnss-sdr/blob/ccd4b17b447b3e905342af9240a05c4d375e3c34/tests/CMakeLists.txt#L541-L553

In general, I’m hesitant to include long input files, even as optional resources, because they must be downloaded at configuration time and this often causes issues in CI environments. Our (limited) solution for long signals has been to rely on the [gnss-simulator](https://bitbucket.org/jarribas/gnss-simulator), see for instance [`position_test`](https://gnss-sdr.org/docs/tutorials/testing-software-receiver-2/#position_test).

Let me know if you need help adapting this for your tests, ideally we should have tests for everything :-)

MathieuFavreau:
@carlesfernandez after all I was able to make a test with a recording of 4ms that was already there.
I had to change the code a little bit to prevent consuming samples between step one and step two, by keeping the `pcps_acquisition` active, otherwise it would not acquire within 4ms of signal.

carlesfernandez:
Update: I've introduced a new CMake helper function [download_test_file()](https://github.com/gnss-sdr/gnss-sdr/blob/654715b60b0880746769013ba545c2361b5720f3/tests/CMakeLists.txt#L35) to fetch optional test signal data, consolidating the previous ad-hoc file(DOWNLOAD ...) logic and making it easier to add new test inputs safely. Thank you for your inspiration :-)

PR #970, opened 2026-01-13
Cleanup RINEX obs

MathieuFavreau:
Simplify a lot the `log_rinex_obs` functions, it should still work for RINEX version 2 and 3 correctly.

Also it will now always populate all the obs values if there are multiple signals for the same constellation, writing 0s for the missing observables, so it will always match the header obs format.

I also fixed an issue that I introduced in my last PR, where the obs epoch record started with two `>` instead of one.

I added the epoch line verification to the tests, and fixed some small errors in the tests which previously made no difference, but now broke the tests.

### DIFF
--- a/src/algorithms/PVT/libs/rinex_printer.cc
+++ b/src/algorithms/PVT/libs/rinex_printer.cc
@@ -57,62 +57,77 @@
 
 namespace
 {
-const std::unordered_map<std::string, std::string> satelliteSystem = {
-    {"GPS", "G"},
-    {"GLONASS", "R"},
-    {"SBAS payload", "S"},
-    {"Galileo", "E"},
-    {"Beidou", "C"},
-    {"Mixed", "M"}};  // RINEX v3.02 codes
+const std::unordered_map<std::string, char> satelliteSystem = {
+    {"GPS", 'G'},
+    {"GLONASS", 'R'},
+    {"SBAS payload", 'S'},
+    {"Galileo", 'E'},
+    {"Beidou", 'C'},
+    {"Mixed", 'M'}};  // RINEX v3.02 codes
+
+
+std::string signal_flag_to_string(signal_flag flag)
+{
+    switch (flag)
+        {
+        case GPS_1C:
+            return "1C";
+        case GPS_2S:
+            return "2S";
+        case GPS_L5:
+            return "L5";
+        case GAL_1B:
+            return "1B";
+        case GAL_E5a:
+            return "5X";
+        case GAL_E5b:
+            return "7X";
+        case GAL_E6:
+            return "E6";
+        case GLO_1G:
+            return "1G";
+        case GLO_2G:
+            return "2G";
+        case BDS_B1:
+            return "B1";
+        case BDS_B3:
+            return "B3";
+        }
+
+    throw std::runtime_error("Invalid signal");
+}
+
+
+std::map<std::string, signal_flag> string_to_signal_flag_map()
+{
+    std::map<std::string, signal_flag> convertion_map;
+
+    for (const auto flag : {GPS_1C, GPS_2S, GPS_L5, GAL_1B, GAL_E5a, GAL_E5b, GAL_E6, GLO_1G, GLO_2G, BDS_B1, BDS_B3})
+        {
+            convertion_map[signal_flag_to_string(flag)] = flag;
+        }
+
+    return convertion_map;
+}
+
+
+signal_flag string_to_signal_flag(const std::string& signal_str)
+{
+    static const auto convertion_map = string_to_signal_flag_map();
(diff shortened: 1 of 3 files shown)


tests/unit-tests/signal-processing-blocks/pvt/rinex_printer_test.cc
```diff
@@ -768,7 +768,7 @@ TEST_F(RinexPrinterTest, MixedObsLog)
     eph_gal.PRN = 1;
     Pvt_Conf conf;
     conf.use_e6_for_pvt = false;
-    const auto signal_enabled_flags = GPS_1C | GAL_1B;
+    const auto signal_enabled_flags = GPS_1C | GAL_1B | GAL_E5a;
```
MathieuFavreau, line 741: The test was writing `E5a` obs

tests/unit-tests/signal-processing-blocks/pvt/rinex_printer_test.cc
```diff
@@ -517,7 +517,7 @@ TEST_F(RinexPrinterTest, GlonassObsLog)
     gs3.System = *sys.c_str();
     gs4.System = *sys.c_str();
 
-    std::string sig = "1C";
+    std::string sig = "1G";
```
MathieuFavreau, line 532: This is the correct signal ID for GLONASS L1

tests/unit-tests/signal-processing-blocks/pvt/rinex_printer_test.cc
```diff
@@ -928,6 +928,8 @@ TEST_F(RinexPrinterTest, MixedObsLogGpsGlo)
     std::memcpy(reinterpret_cast<void*>(gs5.Signal), sig.c_str(), 3);
     std::memcpy(reinterpret_cast<void*>(gs6.Signal), sig.c_str(), 3);
     std::memcpy(reinterpret_cast<void*>(gs7.Signal), sig.c_str(), 3);
+
+    sig = "2G";
```
MathieuFavreau, line 889: This test was writing two obs for `1G`, which makes no sense, so I changed the second one to `2G`

src/algorithms/PVT/libs/rinex_printer.cc
```diff
@@ -129,6 +144,73 @@ std::string enabled_signal_flags_to_string(const Signal_Enabled_Flags& flags)
...
+                            // Check if the signal entry exist
+                            const auto& signal_obs_it = signal_obs_map.find(signal_flag);
+
+                            // If it doesn't, it's the first one and we create an entry for each active signal, otherwise just update
+                            if (signal_obs_it == signal_obs_map.end())
+                                {
+                                    for (const auto cosntel_signal_flag : signal_flags_it->second)
+                                        {
+                                            signal_obs_map[cosntel_signal_flag] = Gnss_Synchro{};
+                                        }
```
MathieuFavreau, line 200: Make sure to create an empty observable for every active signal of that constellation

src/algorithms/PVT/libs/rinex_printer.cc
```diff
@@ -1248,17 +1344,15 @@ void add_svclk_to_line(const boost::posix_time::ptime& utc_time, bool log_second
...
-std::string get_obs_epoch_record_lines(const boost::posix_time::ptime& utc_time, double seconds)
+std::string get_obs_epoch_record_lines(const boost::posix_time::ptime& utc_time, double seconds, int version)
 {
     // -------- EPOCH record
     std::string line = std::string(1, '>');
 
     // double utc_t = nav_msg.utc_time(nav_msg.sv_clock_correction(obs_time));
     // const double gps_t = eph.sv_clock_correction(obs_time);
 
-    line += std::string(1, '>');
```
MathieuFavreau, line 1259: That was the extra `>`

MathieuFavreau:
@carlesfernandez I will have a couple other PR following, my goal is to hopefully remove the necessity to hardcode every possible combination of signals in the RINEX printer

carlesfernandez:
I tested this and a failure occurs specifically in BeiDou observation annotations (B1 and/or B3). I’m investigating the root cause.

carlesfernandez:
Ok, there it is: in Gnss_Synchro.System, BeiDou is represented by 'B' instead of 'C'. That's unfortunate, since the other systems follow the RINEX convention. I'm more in favour of changing 'B' to 'C' in Gnss_Synchro for consistency. I can take care of that.

carlesfernandez:
Testing this: https://github.com/gnss-sdr/gnss-sdr/commit/8ec3c911dbdc588c4e685850ed9b9a07bb2abc17 

If the test pipeline completes successfully, I’ll merge this PR. Great work!

PR #985, opened 2026-02-16
Cleanup rinex nav header

MathieuFavreau:
- Removed SBAS commented code for now
- Moved `getLocalTime` to `.cpp` unnamed namespace
- Add more functions to remove duplication for RINEX v2
- Create generic `rinex_nav_header` function that work for any constellation and RINEX version

Note: sometimes GPS IONO beta was not logged, so I've added it, but I am unsure if there was a reason why it was like that


MathieuFavreau:
The Clang-format CI is broken, see [this issue](https://github.com/jidicula/clang-format-action/issues/267)

PR #986, opened 2026-02-17
Make RINEX nav header writting work with any signal combination

MathieuFavreau:
- Remove the `rinex_nav_header` functions that were per signals, instead build header based on active signals
- Move checks for ephemeris at the start of the RINEX logging function to simplify the code
- Check if `d_rinex_header_updated` is false once first
- Fix RINEX v2 obs file with GPS + GLONASS having duplicated obs header line
- Simplify `update_obs_header`


MathieuFavreau:
One more PR after this one and the RINEX printer will be generic, meaning it won't be necessary anymore to hardcode signal combinations

PR #987, opened 2026-02-18
Make RINEX header update work with any signal combination

MathieuFavreau:
- Move the RINEX update header functions into `print_rinex_annotation` and make it work with any signal combination
- Merged the `log_rinex_nav` public and private functions into one
- Track header updates per constellation

This is the last PR to remove hardcoding of signal combinations from the RINEX printer, it should now work with any signal combination


carlesfernandez:
@MathieuFavreau 

This whole refactoring of the RINEX printer is outstanding work. Many thanks for the effort and attention to detail. Hats off for your contribution! I really love it.

PR #993, opened 2026-03-11
Fix core dump by adding missing channel check in Plutosdr_Signal_Source

bhumikasudarshani-cmd:
### Description of the change
This PR fixes a bug that causes GNSS-SDR to crash with a core dump (Assertion `!d_channels.empty()` failed) if the `SignalSource.channels` parameter is missing or invalid in the configuration file.

I added a safety check in the `Plutosdr_Signal_Source` constructor to verify that `channels >= 1`. If the channels are not configured correctly, it now gracefully logs a `FATAL` error with instructions to set the parameter.

### How to verify it
1. Create a configuration file with `SignalSource.implementation=Plutosdr_Signal_Source` but omit the `SignalSource.channels` parameter.
2. Run GNSS-SDR.
3. Instead of a core dump, the program will cleanly exit with: `Configuration error: SignalSource.channels missing or invalid!`


carlesfernandez:
Please submit your pull request only against the `next` branch.

Additionally, the proposed changes do not seem to implement the behavior you describe.

PR #1002, opened 2026-04-01
Cleanup GNSS block factory

MathieuFavreau:


carlesfernandez:
const correctness is always a good thing 👌 Thank you!

PR #1004, opened 2026-04-15
Fix file config is_present

MathieuFavreau:
@carlesfernandez I was wondering why you needed [in this commit](https://github.com/gnss-sdr/gnss-sdr/commit/59aa060cd059cdbbde2b9c4be2d40ca837bdc6c4) to change the logic and found the issue, `FileConfiguration::is_present` does not read from the file config, only from overridden values.

This means probably a bunch more configurations are currently broken, as I used `is_present` in `get_role_name`, so configs with acquisiton / tracking / telemetry implementations per channels will not work as intended. Could you validate the config that failed and prompted the change you made to my code is now correctly working?

I saw you made a new release yesterday, would it be a lot of work to bring back this bug fix in the release?

### DIFF
--- a/src/core/receiver/file_configuration.cc
+++ b/src/core/receiver/file_configuration.cc
@@ -188,5 +188,10 @@ void FileConfiguration::set_property(std::string property_name, std::string valu
 
 bool FileConfiguration::is_present(const std::string& property_name) const
 {
-    return (overrided_->is_present(property_name));
+    if (overrided_->is_present(property_name))
+        {
+            return true;
+        }
+
+    return ini_reader_->HasValue("GNSS-SDR", property_name);
 }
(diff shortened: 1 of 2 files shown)


carlesfernandez:
If there is some smelly bug around, as it seems, we should fix it and release another version of GNSS-SDR. I honestly was surprised when I commited  59aa060cd059cdbbde2b9c4be2d40ca837bdc6c4, since your code change looked good to me at first sight but, when applied, it broke several CI tests. Reverting it back fixed all tests cases.

There were other unrelated issues (such as recent Boost C++ Libraries API changes, updates to Debian packaging, plus the consolidation of your work across other parts of the codebase) that urged us to publish a new release, but nothing prevents us from issuing a follow-up release if a relevant flaw is identified and fixed.

carlesfernandez:
Well, as usual, you were right and identified the problem correctly. My mistake for committing an incorrect fix. I will merge this into both the `next` and `main` branches, and apply a patch to the GNSS-SDR packaging to minimize the impact of this issue.

A new GNSS-SDR release should come soon.

PR #1014, opened 2026-05-11
Add option to compute both acquisition steps on same samples

MathieuFavreau:
The core of this PR is to change `make_2_steps` to allow both steps on the same samples (using `use_same_samples_on_step_2`).

The second step is usually much less computation heavy, so it doesn't change much the duration of the acquisition call to do both steps at once (2-3% increase from some tests I did).

Other important points:
- Changed `max_dwells` to be `1` when `bit_transition_flag` is `true`, instead of checking `bit_transition_flag` directly in acquisition, for more generic code
- Set `d_gnss_synchro->Acq_doppler_step` even when  `d_step_two` is `false`
- Improved logging to log second step doppler parameters when on second step
- Simplified state logic to reduce number of if branches and make it easier to understand the code

### DIFF
--- a/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.cc
+++ b/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.cc
@@ -195,6 +195,11 @@ Acq_Conf get_acq_conf(const ConfigurationInterface* configuration, const std::st
     acq_parameters.sampled_ms = sig_info.ms_per_code;  // Set as default value
     acq_parameters.SetFromConfiguration(configuration, role, sig_info.chip_rate, sig_info.opt_freq);
 
+    if (acq_parameters.bit_transition_flag)
+        {
+            acq_parameters.max_dwells = 1;  // Max one dwell with bit_transition_flag
+        }
+
     if (sig_flag == GAL_1B)
         {
             acq_parameters.acquire_pilot = configuration->property(role + ".acquire_pilot", acq_parameters.acquire_pilot);
--- a/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
+++ b/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
@@ -300,9 +300,9 @@ void pcps_acquisition::update_grid_doppler_wipeoffs_step2()
 }
 
 
-void pcps_acquisition::log_acquisition(const AcquisitionResult& result) const
+void pcps_acquisition::log_acquisition(const AcquisitionResult& result, bool positive_acq)
 {
-    DLOG(INFO) << (result.positive_acq ? "positive" : "negative") << " acquisition"
+    DLOG(INFO) << (positive_acq ? "positive" : "negative") << " acquisition"
                << ", satellite " << d_gnss_synchro->System << " " << d_gnss_synchro->PRN
                << ", sample_stamp " << result.sample_count
                << ", test statistics value " << result.test_statistics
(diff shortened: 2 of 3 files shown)


MathieuFavreau:
@carlesfernandez do you see a problem with this PR?

MathieuFavreau:
@carlesfernandez I've changed the PR to now be an option in the acquisition config, by default it will be the same behavior as before

vladisslav2011:
Hello.
I've tested this PR alone and in combination with https://github.com/gnss-sdr/gnss-sdr/pull/1095 (to confirm, that both acquisition steps are computed on the same set of samples - the estimated code phase was expected to be always exactly zero even in non-blocking mode)
And it does not seem to work correctly...
1. No buffering is implemented (conflicting PR, that removed buffering is https://github.com/gnss-sdr/gnss-sdr/pull/1101), so when the number of non-coherent integrations is larger than 1, everything goes the wrong way. The first integration of step 2 is computed using the buffer from last integration of step 1 instead of expected buffer from first integration of step 1, then `d_step_two` gets overwritten and so on.
2. New configuration option `full_grid_search` breaks acquisition completely: no satellites are found
3. https://github.com/gnss-sdr/gnss-sdr/pull/1095 does not seem to work well with this PR without some modifications (don't worry, I'll update it if this PR will be merged)
4. It seems to work as expected only when `bit_transition_flag`=true or `max_dwells`=1 and `full_grid_search`=false

You may want to take a look at https://github.com/gnss-sdr/gnss-sdr/pull/1108, as it brings the buffering back and makes it possible to implement the idea of this PR correctly (support non-blocking mode and perform the second step using exactly the same samples, that were used at step 1).

MathieuFavreau:
@vladisslav2011 you are right `use_same_samples_on_step_2` does not work with `max_dwells > 1` within this PR. 
I have disabled for now `use_same_samples_on_step_2` when `max_dwells > 1`, and suggest I fix it in another PR.

From my understanding that is the only concern you had right?

Since my PR simplifies quite a bit the result check state logic (see `check_result`), I think it would be good to merge it first as it could simplify the logic within #1108 which is getting complex.

vladisslav2011:
> From my understanding that is the only concern you had right?

https://github.com/MathieuFavreau/gnss-sdr/blob/bug/make-two-step-in-one-go/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc#L835
This assignment breaks the second step as `d_step_two` gets reset after first non-coherent integration of step 2...

MathieuFavreau:
@vladisslav2011 it should be fixed now

vladisslav2011:
Looks good. I'll pull and retest.

MathieuFavreau:
@vladisslav2011 you had time to retest?

vladisslav2011:
I was a bit busy trying to make https://github.com/gnss-sdr/gnss-sdr/pull/1108 stop crashing/freezing during termination, but now it looks like fixed (may be still not ready for merging as random CI failures are still happening on github machines, but not on my machines).

Let's test this PR now and give https://github.com/gnss-sdr/gnss-sdr/pull/1108 a very long run a bit later.

vladisslav2011:
- [x] Non-blocking mode `max_dwells=20`
- [x] Blocking mode `max_dwells=20`
- [x] Non-blocking mode  `bit_transition_flag=true`
- [x] Blocking mode  `bit_transition_flag=true`
- [x] Non-blocking mode `full_grid_search=true`
- [x] Blocking mode `full_grid_search=true`
- [x] `make_two_steps=false` and different combinations of other options
- [x] `GNSS-SDR.use_acquisition_resampler=true` and different combinations of other options

Everything looks fine now.

vladisslav2011:
@MathieuFavreau could you take a look at https://github.com/gnss-sdr/gnss-sdr/pull/1123
I don't know, how to explain and name it better, so suggestions and questions are welcome.

PR #1023, opened 2026-06-13
WIP: Initial GPS L1C support

tatjam:
**Work in progress** PR to implement GPS L1C support. Creating the PR so I can share progress in the comments. I will switch the PR to ready for review once I'm satisfied with the whole signal chain and written / automated the tests.

I've decided to use "GPS L1" for the L1C signal "code". I'm not sure if this is appropriate, and "GPS L1C" would be better, but I saw that all signals are given 2 letter codes.

## D ranging code acquisition and tracking

**Progress**: Python range code generator and navigation bit extraction with tracker tracking pilot.

The signal replica for the D ranging code is implemented from a look-up table generated by a Python script from the data in the IS-GPS-800J SIS specification.

## P ranging code acquisition and tracking

**Progress**: Python range code generator, PCPS acquisition (without overlay) and VEML tracking (with overlay code)

The GPS L1C pilot code is very long, as it's formed by a 10230 chip ranging code overlayed by a 1800 chip code, for a total length of 18.414Mchips. The overlay code fits nicely into the secondary code function of the VEML tracker, and it can be locked in 18 seconds nominally.

## Telemetry decoder

**Progress**: TOI decoder (following the IS-GPS-800J algorithm) and LDPC decoder, with look up tables generated in Python. Frames are decoded successfully, but no data extraction from them yet.

The LDPC decoder is an implementation of the "Attenuated Min-Sum" decoding algorithm detailed in _"Channel Codes: Classical And Modern"_, William E. Ryan and Shu Lin, 2009. This [diagram](https://github.com/user-attachments/files/28934133/LDPC.pdf) may be helpful to understand the implementation.

## PVT

**Progress**: Ephemeris and clock corrections extracted, other subframes not yet used for PVT.

## Testing dataset 

For testing, i'm using the following dataset: https://etsin.fairdata.fi/dataset/63f8b776-680b-4c98-ace7-d5e443f2b1c5 
with the following configuration file https://gist.github.com/tatjam/5a4574d0c2f4c4c1e5fb7820f5121673  (I will keep the gist updated as I make changes!)


tatjam:
Acquisition and tracking of PRN 5 signal in the dataset using only the P code (without overlay)
image
image

minhaj6:
nice! Really looking forward to this. I was thinking about working on this, but never got around to it. Thanks!!

tatjam:
Welp after a productive weekend of coding I think i've got the telemetry decoding to work! The LDPC eventually finds a syndrome for the code at a sane TOI value:

image

I attach a small diagram I made to help myself implement the LDPC decoding algorithm, and that's useful to the documentation in the LDPC decoder source code. [LDPC.pdf](https://github.com/user-attachments/files/28934133/LDPC.pdf)

Now, the navigation messages as of last commit are decoded (I haven't really checked if correctly, but getting a match from an LDPC with wrong bytes is rare (?)), but nothing useful extracted from them, that's the next step. 

One little caveat of how the code works is that it requires the pilot code for alignment, but I think in the absence of this code, the TOI could be used as a sort of preamble to trigger LDPC decoding attempts until a match is found. 

The TOI alone cannot be used as a preamble (even if it's highly redundant, encoding 9 bits in 52) because it's a cyclic code and thus under small bit shifts it gives nonsense, yet highly correlated, values. I also tried using two subsequent TOI codes, but it also has the cyclic property on subsequent symbols, so you end up getting high correlation peaks at many sequences of TOIs, without a clear winner.

I'm not sure in what situation will a receiver detect the data code and not the secondary one, but maybe it's useful for lower processing power usage?

tatjam:
(Rebased into `next` and fixed conflicts)

carlesfernandez:
This is very cool. Could you please sign your commits?

For previous ones, you can do (from your branch):

```
git fetch upstream
git rebase --signoff upstream/next
git push --force-with-lease
```

src/algorithms/libs/gps_l1c_signal_replica.cc
```diff
@@ -0,0 +1,218 @@
...
+    const int32_t prn_ = prn - 1;
+    int32_t index = 0;
+
+    if ((prn < 1) || (prn > GPS_L1C_NUMBER_OF_PRN))
+        {
+            return "";
+        }
+
+    std::string out{};
+    out.reserve(GPS_L1C_O_CODE_LENGTH_CHIPS);
```
carlesfernandez: This should be `resize()`, not `reserve()`. `reserve()` only changes capacity; the string size remains zero, so the later writes through `out[index]`, `out[index + 1]`, etc. are out-of-bounds and invoke undefined behavior. 

```suggestion
    out.resize(GPS_L1C_O_CODE_LENGTH_CHIPS);

File: src/algorithms/telemetry_decoder/gnuradio_blocks/gps_l1c_telemetry_decoder_gs.h [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+#ifndef GNSS_SDR_GPS_L1C_TELEMETRY_DECODER_GS_H
+#define GNSS_SDR_GPS_L1C_TELEMETRY_DECODER_GS_H
+
+#include "GPS_L1C.h"
+#include "gnss_synchro.h"
+#include "ldpc_min_sum_decoder.h"
+#include "telemetry_impl_interface.h"
+#include <boost/circular_buffer.hpp>
+#include <boost/crc.hpp>
+#include <bitset>
```
carlesfernandez, line 32: This header uses `std::array` directly but does not include `<array>`. It also depends on `std::vector` through template declarations, currently via indirect includes. Could we make the header self-contained by including the standard headers it uses directly?

src/algorithms/telemetry_decoder/gnuradio_blocks/gps_l1c_telemetry_decoder_gs.cc
```diff
@@ -0,0 +1,460 @@
...
+    auto e = static_cast<uint64_t>(extract_unsigned(sf2, 173, 206));
+    auto w = static_cast<int64_t>(extract_signed(sf2, 206, 239));
+    auto Omega = static_cast<int64_t>(extract_signed(sf2, 239, 272));
+    auto i = static_cast<int64_t>(extract_signed(sf2, 272, 305));
+    auto delta_dot_Omega = static_cast<int32_t>(extract_signed(sf2, 305, 322));
+    auto idot = static_cast<int16_t>(extract_signed(sf2, 322, 337));
+    auto cisn = static_cast<int16_t>(extract_signed(sf2, 337, 353));
+    auto cicn = static_cast<int16_t>(extract_signed(sf2, 353, 369));
+    auto crsn = static_cast<int32_t>(extract_signed(sf2, 369, 393));
+    auto crcn = static_cast<int32_t>(extract_signed(sf2, 394, 417));
```
carlesfernandez: This extraction appears to skip bit 393: `crsn` extracts `[369, 393)`, then `crcn` starts at `394`. If these fields are intended to be contiguous and `crcn` is a 24-bit signed field like the surrounding harmonic terms, this should likely start at `393`.

```suggestion
    auto crcn = extract_signed(sf2, 393, 417);

carlesfernandez (commented):
Thanks for pushing initial GPS L1C support. I reviewed this as WIP and found two blockers worth fixing before deeper validation: the overlay-code generator currently writes out of bounds after `reserve()`, and a likely SF2 bit-offset issue.

tatjam:
Thanks for the review! I've fixed the comments, and also ran clang-format on the files that the checks was marking. 

I've started refactoring the navigation data bit extractor to be more similar to the existing one for GPS CNAV to avoid excessive code duplication / having two approaches for the same thing. I'll move the bit extraction logic to a Gps_CNAV2_Navigation_Message and use bitsets and such.

Similarly, for the CNAV2 Ephemeris (and possibly Almanac, iono, clock correction...) I think the best approach is to create a new class that inherits from the CNAV one and just add the new inter-signal correction terms there, instead of duplicating all the code. Afterwards, the integration into the PVT solver should be fairly straightforward.

src/algorithms/telemetry_decoder/gnuradio_blocks/gps_l1c_telemetry_decoder_gs.h
tatjam: I'll refactor this code completely to make it more similar to existing GPS CNAV navigation message parsing.

src/algorithms/telemetry_decoder/gnuradio_blocks/gps_l1c_telemetry_decoder_gs.cc
tatjam: Same as before, this will be refactored, but fixed it so when I move the bit offsets to constants I don't make the same mistake  again.

tatjam:
PVT is not yet working, but the navigation message is now parsed in a similar way to how the CNAV one is. I've also refactored the "read_navigation_unsigned", "read_navigation_signed" and "read_navigation_bool" into a separate header file, as they were duplicated in the navigation decoders, and would be duplicated once more by the CNAV-2 decoder.

tatjam:
Finally, PVT is working! In green the reference position from the dataset, in blue the samples generated by gnss-sdr:

image

Now I'll try to fix the various TODOs I have around, and also write some unit tests for the various components.

PR #1025, opened 2026-06-19
Improved plot_acq_grid.py. Fixed python tracking dump reader.

minhaj6:
- Added a plot_positive_acq filter option, to enable plotting only the positive acquisitions.
- Improved the file path processing. Previous version used hardcoded character offset that will not work if someone sets custom dump prefix. Now we use python's built-in pathlib and split-parse using the "_" separator.
- The scripts used to use two different ways of encoding system,channel,chips, one with python dict another with massive if-else nest. This has been uniformed.
- The script used to assume all files in the dump folder would be acquisition dump but now we filter for acq dumps using pathlib's glob()

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@
 .settings
 .vs/
 .vscode/
+*.pyc
--- a/utils/python/gps_l1_ca_kf_plot_sample.py
+++ b/utils/python/gps_l1_ca_kf_plot_sample.py
@@ -1,17 +1,21 @@
 """
  gps_l1_ca_kf_plot_sample.py
 
- Reads GNSS-SDR Tracking dump binary file using the provided
- function and plots some internal variables
+ Reads a GPS L1 C/A Kalman-filter tracking dump binary file
+ (GPS_L1_CA_KF_Tracking) and plots some internal tracking and Kalman-filter
+ variables.
 
  Irene Pérez Riega, 2023. iperrie@inta.es
+ Minhaj Uddin Ahmad, 2026. mahmad12@crimson.ua.edu
 
  Modifiable in the file:
-   sampling_freq     - Sampling frequency [Hz]
-   channels          - Number of channels to check if they exist
+   samplingFreq      - Sampling frequency [Hz]
+   channels          - Number of channels
    first_channel     - Number of the first channel
-   path              - Path to folder which contains raw file
-   'trk_dump_ch'     - Fixed part in tracking dump files names
+   code_period       - Code period [s]
+   path              - Path to folder which contains the tracking dump files
+   fig_path          - Path where the plots will be saved
+   file_prefix       - Fixed part of the tracking dump file names
 
  -----------------------------------------------------------------------------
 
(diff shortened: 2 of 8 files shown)


minhaj6:
The Acquisition block writes a dump whenever the dwell budget is exhausted, regardless of detection - `dump_results()` being called on the `max_dwells` condition in `pcps_acquisition.cc:719`. 

This produces many `positive_acq = 0` files; `plot_positive_acqs` lets users plot only the actual detections.

minhaj6:
```
minhaj@thonk-carbon:~/GPS-Data/PocketSDR-Recordings/gnss-sdr/utils/python$ python plot_tracking_quality_indicators.py 
Traceback (most recent call last):
  File "/home/minhaj/GPS-Data/PocketSDR-Recordings/gnss-sdr/utils/python/plot_tracking_quality_indicators.py", line 44, in <module>
    GNSS_tracking.append(dll_pll_veml_read_tracking_dump(tracking_log_path))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/minhaj/GPS-Data/PocketSDR-Recordings/gnss-sdr/utils/python/lib/dll_pll_veml_read_tracking_dump.py", line 104, in dll_pll_veml_read_tracking_dump
    v7.append(struct.unpack('f',
              ^^^^^^^^^^^^^^^^^^
struct.error: unpack requires a buffer of 4 bytes
```
The latest commit fixes the broken Python tracking dump reader.

minhaj6:
Same story with the `utils/python/gps_l1_ca_kf_plot_sample.py` script. Initially I couldn't tell the difference between them two readers, as I didn't find any reference to KF in the docs - https://gnss-sdr.org/docs/sp-blocks/tracking/ . 
But looks like some kf based gnu-radio blocks/adapters are present in the `src/algorithms/tracking/`. 

Not sure if it was an experimental feature and never made it to the docs. 🤔

carlesfernandez:
Thanks a lot for this contribution, @minhaj6!  Improving the Python plotting utilities is genuinely useful, and your work on the acquisition grid script was the spark for a broader cleanup.

Building on your changes, I've extended the parameterization across the rest of the scripts in utils/python/ so that users no longer need to edit the source code to use them:

- Every script now has a proper command-line interface (--help) and can be run from any directory, with configurable input and output locations.
- A --file-prefix option takes the value of the corresponding block's dump_filename configuration parameter directly, reconstructing the dump filenames the same way the receiver does.
- A --format option selects the output image format (png by default, plus pdf, eps, svg, jpg), with the vector formats written publication-ready with embedded fonts — handy for papers and reports.
- Added a utils/python/README.md documenting all the utilities and their options.

Really appreciate you getting this started — great to have these tools more user-friendly.

minhaj6:
Thanks @carlesfernandez for working on this! That's a better design. 
Eventually, maybe we can expose some utilities to the path (e.g., make install), so people can access them from anywhere. Just throwing an idea; let me know your thoughts 🤔 

Something like this - 
image

PR #1028, opened 2026-06-29
pcps_acquisition: fix crash in PcpsAcquisitionAdapterCustom::set_local_code

vladisslav2011:
... when Acquisition_5X.coherent_integration_time_ms is greater than 1. In this case vectors codeQ_, codeI, codeQ are constructed with code_length size, but written with vector_length which results in arbitrary memory overwriting and crash.
Construct these arrays with vector_length size to make sure that there will be enough space for all primary code replicas and zero padding.


carlesfernandez:
Thanks! Merging

PR #1030, opened 2026-06-29
QZSS: fix crash in RTCM generator

vladisslav2011:
xxx_last_lock_time arrays are limited to 64 elements but QZSS PRN number
may be as high as 206. This leads to a segmentation fault as a result of
out-of-bounds access.
Resize xxx_last_lock_time to 207 elements and remove magic numbers
"64"/"65" in array index calculations.
Add private static field Rtcm::PRN_MAX to define maximum possible PRN
number.

### DIFF
--- a/src/algorithms/PVT/libs/rtcm.cc
+++ b/src/algorithms/PVT/libs/rtcm.cc
@@ -2558,10 +2558,10 @@ std::string Rtcm::get_MSM_1_content_sat_data(const std::map<int32_t, Gnss_Synchr
         gnss_synchro_iter != observables.cend();
         gnss_synchro_iter++)
         {
-            it = std::find(pos.begin(), pos.end(), 65 - gnss_synchro_iter->second.PRN);
+            it = std::find(pos.begin(), pos.end(), PRN_MAX + 2 - gnss_synchro_iter->second.PRN);
             if (it == pos.end())
                 {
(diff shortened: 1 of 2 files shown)


carlesfernandez:
Thank you for tracking this down and for proposing a fix. I agree with the root problem: QZSS PRNs can currently reach RTCM code paths that were written around GPS/Galileo/GLONASS satellite IDs, and that can lead to invalid indexing and crashes.

I am going to reject this PR, though, because the proposed solution is not aligned with RTCM 3.2. QZSS MSM messages are defined separately as message types 1111-1117, and QZSS PRNs 193-202 must be mapped to RTCM satellite mask IDs 1-10. Enlarging the lock-time arrays to accept raw QZSS PRNs does not produce valid RTCM output, and the new indexing formula also risks an out-of-bounds access for ordinary PRN 1 satellites.

Instead, I am applying a narrower defensive fix for now: QZSS data will be skipped in the current RTCM GPS/MSM output paths until proper QZSS RTCM support is implemented, and the lock-time indexing is hardened with bounds checks. This avoids the crash while keeping the generated RTCM messages compliant with the subset currently implemented. See b898bcad28000962ca62fcfed13eeb0894a1326d

A future PR adding proper QZSS RTCM support, including message 1044 and MSM 1111-1117 with the correct PRN-to-satellite-ID mapping, would be very welcome.

Issue #1035, opened 2026-06-29
QZSS: tracking L1C/B signal fails


vladisslav2011:
Steps to reproduce:
1. Setup GNSS-SDR to acquire and track some QZSS satellites at L1 band using hardware signal source
```
Channels_J1.count=4
Acquisition_J1.implementation=QZSS_L1_PCPS_Acquisition
...
```
2. Wait for QZS1R to become visible using, for example, Gpredict
3. Run GNSS-SDR

Expected behavior:
PRN203 is acquired and tracked.
Telemetry messages from QZS1R are being decoded.

Observed behavior:
Failed acquisitions

I've attempted to fix this by enabling acquisition of PRN203...206, extending QZSS_L1_CODE_LENGTH to 2046, changing QZSS_L1_CHIP_RATE  to 2.046e6 and implementing BOC(1,1) generation for L1C/B PRNs in qzss_l1_code_gen_float and generation of duplicate symbols for L1C/A PRNs (to keep L1C/A tracking/acquisition working) but got nothing. I'm out of ideas.

carlesfernandez:
Hi @vladisslav2011 ,

according to [this page](https://sys.qzss.go.jp/dod/en/constellation.html) the QZS1R satellite is NOT broadcasting QZSS L1 C/A data. L1C/B is still not implemented.

vladisslav2011:
I've implemented L1C/B signal support and tested it both in a realtime input and file input configurations.
Both QZS1R (PRN203) and QZS06 (PRN205) may be tracked and used in a solution.
Acquisition requires relatively small doppler step: 20...25 Hz or so and may take a little bit long unless there are 5...10 channels in acquisition and 8...10 channels allocated for QZSS reception. This configuration is not optimal as there are only 5 active satellites and (in my QTH) maximum number of usable visible satellites is 3. A good frequency reference (TCXO or OCXO) is highly recommended to reduce doppler search range and CPU load.
PR will be soon.

vladisslav2011:
PR ready: https://github.com/gnss-sdr/gnss-sdr/pull/1036

PR #1036, opened 2026-07-03
QZSS L1 C/B implementation

vladisslav2011:
Extend QZSS L1 primary code length to 2046 and increase L1 code rate 2.046MHz to handle both L1 C/a and L1 C/B codes.
Implement L1 C/B code generation.
Increase QZSS maximum PRN to 206.

Tested using both real world and simulated data.
May close https://github.com/gnss-sdr/gnss-sdr/issues/1035


carlesfernandez:
Hi @vladisslav2011,

Thank you for this contribution, and for the golden-vector test in the second commit. I have verified the L1 C/B code table entries and the generated sequences against IS-QZSS-PNT, including the "first 10 chips" check values, and the implementation is correct. The PR will be accepted. I am adding a follow-up commit on top of yours with some adjustments before merging into next:

* **Chip-rate semantics**. QZSS_L1_CHIP_RATE is restored to 1.023 Mcps and QZSS_L1_CODE_LENGTH to 1023 chips, with the factor of two moved to a new QZSS_L1_SAMPLES_PER_CHIP = 2 constant, which is the same approach used for Galileo E1 (d_code_samples_per_chip = 2). The generated replica is bit-identical to yours (your golden vectors pass unchanged), but ```Tracking_J1.early_late_space_chips``` is expressed in true chips again, consistently with every other signal. Please note that configs tuned for the half-chip units need the spacing halved: for L1 C/B use a value below 0.33 chips (e.g. 0.25), since the BOC(1,1) correlation main peak ends at ±1/3 chip. A warning is now logged otherwise.

* **DLL discriminator for BOC**. When the tracked PRN is L1 C/B, start_tracking() now normalizes the DLL discriminator with the sinBOC(1,1) correlation function, as done for Galileo E1, instead of the BPSK triangle, fixing a ~1.5× discriminator gain mismatch.

* **Satellite identity in output products**. PRNs 203–206 are mapped to the PRN of the satellite's nominal PNT signals (203→196, 204→197, 205→200, 206→201, per RINEX 4.00 Table 6) when observables and ephemerides enter the PVT block. This keeps RINEX and RTCM outputs within the defined QZSS satellite range (J01–J10) and lets rtklib pair L1 C/B and L5 observations of the same physical satellite.

* **Search lists**. PRNs 203–206 are excluded from the J5 acquisition rotation (they have no L5 code assigned), and the reserved, non-operational PRNs 198 and 202 are left out of the default QZSS search list — still selectable via QZSS.prns.

All changes build cleanly and the unit tests pass. Thanks again for a high-quality contribution.

Issue #1043, opened 2026-07-09
NMEA: missing SNR in L2/L3/L5 configurations


vladisslav2011:
Steps to reproduce:

1. Prepare a configuration file (realtime or file input) with GPS L2 or L5 channels and enabled/configured NMEA printer
2. Prepare to run GNSS-SDR along with GPSD (`socat -d -d pty,paw,echo=0 pty,paw,echo=0`, put the first pty into a conf and so on)
3. Run GPSD client (`xgps` for example)
4. Run GNSS-SDR
5. Wait for a PVT

Expected behavior:

xgps shows used satellites with SNR values matching GNSS-SDR console output

<img width="900" height="711" alt="Image" src="https://github.com/user-attachments/assets/9ce4208c-d22f-4ba9-8f62-40f2c50c8da5" />

Observed behavior:

xgps shows zero used satellites with zero SNR, but with valid azimuth/elevation angles.

<img width="900" height="711" alt="Image" src="https://github.com/user-attachments/assets/71f34692-f74a-44ce-b415-1eaf12d1d02f" />

PR #1044, opened 2026-07-09
Fix NMEA missing SNR and used satellites in non-L1 configurations

vladisslav2011:
Use maximum SNR of all received signals instead of hardcoded band 1 signal.
May close https://github.com/gnss-sdr/gnss-sdr/issues/1043
TBD: tests


carlesfernandez:
Thank you for the contribution!

I added a small follow-up on top of your changes to keep the NMEA GSV signal identifier aligned with the frequency that provides the reported C/N0. This preserves the fix for missing SNR in GPS L2/L5 configurations while avoiding cases where a non-L1 C/N0 value could still be emitted with an L1 signal ID.

I also added focused NMEA printer regression tests for the original issue and for mixed-signal GSV output in the`NmeaPrinterTest.*` test suite.

PR #1049, opened 2026-07-21
dll_pll_veml_tracking: smooth transition to narrow

vladisslav2011:
Perform transition to narrow DLL/PLL bandwidth in small steps. This prevents a loss of lock when the narrow tracking loop filter bandwdth is really narrow, 0.5 Hz or less.
Enable extended correlations only after reaching narrow PLL bandwidth.
This PR adds 2 new configuration options:
Tracking_xx.pll_bw_step and Tracking_xx.dll_bw_step. Default value "0.0" disables slow transition.
Set these configuration options to some small value in range of 0.01 ... 0.001 to enable slow transition.


vladisslav2011:
Draft until I'll find a way to test this. Maybe "default position test" with different settings will detect regressions...

PR #1053, opened 2026-07-27
GPS CNAV: fix satellite position computation

vladisslav2011:
Implement satellite positions computation following the procedure,
described on page 166 of IS-GPS-200 (08/01/2022)
Implement L2C/L5I group delay compensation following the procedure,
described on page 175 of IS-GPS-200 (08/01/2022)

My tests are showing noticeable improvement in position accuracy and solution stability, but I'm not sure if I've not missed anything, so it needs a review and some testing using different datasets.

### DIFF
--- a/src/algorithms/libs/rtklib/rtklib_pntpos.cc
+++ b/src/algorithms/libs/rtklib/rtklib_pntpos.cc
@@ -163,7 +163,7 @@ double getiscl5q(int sat, const nav_t *nav)
 
 /* psendorange with code bias correction -------------------------------------*/
 double prange(const obsd_t *obs, const nav_t *nav, const double *azel,
-    int iter, const prcopt_t *opt, double *var)
+    int iter, const prcopt_t *opt, double *var, int *fidx)
 {
     const double *lam = nav->lam[obs->sat - 1];
     double PC = 0.0;
@@ -296,6 +296,7 @@ double prange(const obsd_t *obs, const nav_t *nav, const double *azel,
                 {
                     P1 += P1_C1; /* C1->P1 */
                     PC = P1 - P1_P2;
+                    *fidx = i;
                 }
             else if (obs->code[i] == CODE_NONE && obs->code[j] != CODE_NONE)
                 {
@@ -305,17 +306,20 @@ double prange(const obsd_t *obs, const nav_t *nav, const double *azel,
                             // PC = P2 - gamma_ * P1_P2 / (1.0 - gamma_);
                             if (obs->code[j] == CODE_L2S)  // L2 single freq.
                                 {
-                                    PC = P2 + P1_P2 - ISCl2;
+                                    PC = P2 - P1_P2 + ISCl2;
+                                    *fidx = j;
                                 }
                             else if (obs->code[j] == CODE_L5X)  // L5 single freq.
                                 {
-                                    PC = P2 + P1_P2 - ISCl5i;
+                                    PC = P2 - P1_P2 + ISCl5i;
+                                    *fidx = j;
                                 }
                         }
                     if (sys == SYS_BDS)
                         {
                             P2 += P2_C2; /* C2->P2 */
                             PC = P2;     // no tgd corrections for B3I
+                            *fidx = j;
                         }
                     else if (sys == SYS_GAL)
                         {
@@ -323,11 +327,13 @@ double prange(const obsd_t *obs, const nav_t *nav, const double *azel,
                             // Galileo OS SIS ICD, Eq. 19: E5a/E5b uses
                             // (f_E1/f_E5)^2 times its corresponding BGD.
                             PC = uses_galileo_bgd ? P2 - gamma_ * P1_P2 : P2 - gamma_ * P1_P2 / (1.0 - gamma_);
+                            *fidx = j;
                         }
                     else if (sys == SYS_GLO)
                         {
                             P2 += P2_C2; /* C2->P2 */
                             PC = P2 - gamma_ * P1_P2 / (1.0 - gamma_);
+                            *fidx = j;
                         }
                 }
             /* dual-frequency */
@@ -477,6 +483,7 @@ int rescode(int iter, const obsd_t *obs, int n, const double *rs,
     int nv = 0;
     int sys;
     int mask[4] = {0};
+    int fidx = 0;
 
     trace(3, "resprng : n=%d\n", n);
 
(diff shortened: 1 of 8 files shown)


src/algorithms/libs/rtklib/rtklib_pntpos.cc
```diff
@@ -477,6 +483,7 @@ int rescode(int iter, const obsd_t *obs, int n, const double *rs,
     int nv = 0;
     int sys;
     int mask[4] = {0};
+    int fidx = 0;
```
carlesfernandez, line 487: `fidx` is initialized only once before the observation loop, but `prange()` does not assign it on every successful path. For example, an L2-only satellite followed by an L1+L2 satellite leaves `fidx == 1` even though the second pseudorange selects L1. The ionospheric correction is then scaled using the L2 wavelength, producing observation-order-dependent errors. Could we initialize the output at `prange()` entry and set it explicitly on every successful branch?

src/core/system_parameters/gnss_ephemeris.h
```diff
@@ -67,23 +67,25 @@ class Gnss_Ephemeris
...
-    double Cus{};       //!< Amplitude of the sine harmonic correction term to the argument of latitude [rad]
-    double Crc{};       //!< Amplitude of the cosine harmonic correction term to the orbit radius [meters]
-    double Crs{};       //!< Amplitude of the sine harmonic correction term to the orbit radius [meters]
-    double Cic{};       //!< Amplitude of the cosine harmonic correction term to the angle of inclination [rad]
-    double Cis{};       //!< Amplitude of the sine harmonic correction term to the angle of inclination [rad]
-    int32_t toe{};      //!< Ephemeris reference time [s]
+    uint32_t PRN{};       //!< SV ID
+    double M_0{};         //!< Mean anomaly at reference time [rad]
+    double Adot{};        //!< Change rate in semi-major axis (CNAV)
+    double delta_ndot{};  //!< Rate of mean motion difference from computed value (CNAV)
```
carlesfernandez, line 73: `delta_ndot` is now used during orbit propagation, but `Gps_CNAV_Ephemeris::serialize()` does not archive it. I confirmed that an XML round trip preserves `Adot` but restores `delta_ndot` as zero, so CNAV ephemerides loaded from storage or assistance XML produce different positions. Please add versioned serialization for this field and bump `BOOST_CLASS_VERSION`.

src/core/system_parameters/gnss_ephemeris.cc
```diff
@@ -190,8 +190,11 @@ void Gnss_Ephemeris::satellitePosVelComputation(double transmitTime, std::array<
     // Time from ephemeris reference epoch
     double tk = check_t(transmitTime - static_cast<double>(this->toe));
 
+    // Semi-major axis correction (CNAV)
+    const double Ak = a + this->Adot * tk;
+
     // Corrected mean motion
-    const double n = n0 + this->delta_n;
+    const double n = n0 + this->delta_n + 0.5 * this->delta_ndot * tk;
```
carlesfernandez: This value is the interval-average mean motion needed for `M = M0 + n * tk`, but it is later reused as the instantaneous rate in `ekdot`. Differentiating the mean anomaly gives `n0 + delta_n + delta_ndot * tk`, without the `0.5` factor. With valid-range synthetic CNAV values at `tk = 7200 s`, the returned velocity differed from a central finite difference of the returned positions by 4.78 m/s. Could we keep separate variables for mean-anomaly propagation and its instantaneous derivative?

src/core/system_parameters/gnss_ephemeris.cc
```diff
@@ -240,8 +243,8 @@ void Gnss_Ephemeris::satellitePosVelComputation(double transmitTime, std::array<
     const double ukdot = pkdot * (1.0 + 2.0 * (this->Cus * c2pk - this->Cuc * s2pk));
 
     // Correct radius
-    const double r = a * OneMinusecosE + this->Crc * c2pk + this->Crs * s2pk;
-    const double rkdot = a * this->ecc * sek * ekdot + 2.0 * pkdot * (this->Crs * c2pk - this->Crc * s2pk);
+    const double r = Ak * OneMinusecosE + this->Crc * c2pk + this->Crs * s2pk;
+    const double rkdot = this->Adot * (1. - this->ecc * cek) + a * this->ecc * sek * ekdot + 2.0 * pkdot * (this->Crs * c2pk - this->Crc * s2pk);
```
carlesfernandez: Since the corrected radius uses `Ak * (1 - e * cos(E))`, its derivative requires `Ak * e * sin(E) * ekdot`, not `a * e * sin(E) * ekdot`. Using `a` leaves the analytical velocity inconsistent with the propagated position whenever `Adot != 0`. Please use `Ak` in this term.

carlesfernandez (commented):
Thanks for this correction. The new L2C and L5I group-delay signs agree with IS-GPS-200N and IS-GPS-705J, and the affected libraries compile successfully.

I found four correctness issues in the new frequency selection and CNAV propagation paths; details are included inline.

- Selected-frequency state can leak between observations.
- `delta_ndot` is lost during XML serialization.
- The eccentric-anomaly rate uses average rather than instantaneous mean motion.
- The corrected radius rate uses `a` instead of `Ak`.

Focused regression tests for mixed-frequency selection, CNAV XML round trips, and analytical velocity versus finite differences would also be valuable.

src/algorithms/libs/rtklib/rtklib_pntpos.cc
vladisslav2011: Done.
The only question is: which frequency index to choose in case of dual (triple) frequency observation and iono_model not equal to "Iono-Free-LC"?
It is kept at zero now, but is it correct?

src/core/system_parameters/gnss_ephemeris.h
vladisslav2011: Thanks. I've missed this. Corrected.

src/core/system_parameters/gnss_ephemeris.cc
vladisslav2011: Updated.
+ Updated mean anomaly calculation in `sv_clock_relativistic_term`
The `0.5` factor was copied from the ICD, so it may be part of `delta_ndot` scaling, not included into LSB from table 30-I.

src/core/system_parameters/gnss_ephemeris.cc
vladisslav2011: Thanks. Updated.

vladisslav2011:
`ekdot` calculation  update and some tests.

>and analytical velocity versus finite differences would also be valuable

Implemented GPS, Galileo, BeiDou predicted doppler, satellite position/velocity and analytical velocity versus finite differences regression tests. No Glonass so far, as nobody have implemented `Glonass_Gnav_Ephemeris::satellitePosition` and `Glonass_Gnav_Ephemeris::predicted_doppler`.

PR #1058, opened 2026-07-30
Add optional median-based outlier pseudorange rejection.

OuWenhao16:
Gate the check behind Observables.reject_outlier_pseudoranges (default false) so existing configs keep previous behavior. When enabled, mark pseudoranges far from the epoch median as invalid.

However, this function is currently worthy of discussion and verification.

### DIFF
--- a/src/algorithms/observables/adapters/hybrid_observables.cc
+++ b/src/algorithms/observables/adapters/hybrid_observables.cc
@@ -51,6 +51,7 @@ HybridObservables::HybridObservables(const ConfigurationInterface* configuration
     conf.always_output_gs = configuration->property("PVT.an_output_enabled", conf.always_output_gs) || configuration->property(role + ".always_output_gs", conf.always_output_gs);
     conf.enable_E6 = configuration->property(role + ".enable_E6", conf.enable_E6);
     conf.enable_monitor = configuration->property("Monitor.enable_monitor", conf.enable_monitor);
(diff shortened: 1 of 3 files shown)


minhaj6:
>  worthy of discussion 

I am a bit curious about this decision and why not RAIM (already implemented). 

Thank you!

OuWenhao16:
> > worthy of discussion
> 
> I am a bit curious about this decision and why not RAIM (already implemented).
> 
> Thank you!

RAIM FDE is already available via `PVT.raim_fde` (default off) and acts inside `pntpos` after a failed single-point solution.  

This change is complementary and sits earlier, in the observables stage. When adding or debugging new signal paths, TOW alignment / interpolation issues can occasionally produce grossly wrong pseudoranges before PVT. Those can spoil an epoch in ways RAIM is not primarily meant to guard against.  This was also a problem I encountered when attempting to increase the B1C frequency band. I think adding this switch will help in troubleshooting such issues. However, this is just my personal opinion. If there are other better ways to solve the problem, this feature might indeed not be necessary, so I also mentioned "currently worthy of discussion".

The median check is therefore an optional gross-error gate (`Observables.reject_outlier_pseudoranges`, default false, large threshold), not a replacement for RAIM. For normal operation RAIM (and the existing chi-square / GDOP checks) remain the appropriate integrity tools.

PR #1061, opened 2026-08-03
CMake: fix building against modern protobuf

l29ah:
tested with protobuf 34.2 on Gentoo
the Debug check was dropped altogether as it breaks the same on RelWithDebInfo


carlesfernandez:
Thanks for spotting this issue. Please let me know if the pushes made to upstream's `next` fix it for you.

l29ah:
The current `next` builds and runs, thanks.

PR #1074, opened 2026-08-27
Fix tracking-dump I/Q swap and Python plot channel/signal mislabeling

joebre:
## Summary
Three small, independent fixes found while validating a dual-band (Galileo E1+E5a) tracking-dump plotting workflow:

- **C++**: `dll_pll_veml_tracking::log_data()` read the data-arm Prompt correlator's real/imaginary components directly instead of going through `map_correlator_to_iq()`, unlike `assign_correlators_to_synchro()` which already does this correctly. For pilot-tracked signals whose data and pilot components are true RF quadrature (Galileo E5a/E5b, GPS L5), `d_interchange_iq` is `true` and the data bits are on the imaginary axis of `d_Prompt_Data`, not the real one. This only affects the `.dat`/`.mat` tracking dump used by the plotting utilities — receiver operation (telemetry decoding, PVT) is unaffected since it uses the already-correct `assign_correlators_to_synchro()` path.
- **Python**: `utils/python/lib/gnss_sdr_conf.py`'s `SIGNAL_ORDER` (meant to mirror `signal_mapping` in `gnss_block_factory.cc`, used to infer channel-ID ranges per signal type) was missing `"1D"` (BeiDou B1C) and `"S1"` (SBAS L1).
- **Python**: `utils/python/dll_pll_veml_plot_sample.py`'s `apply_conf_defaults()` mislabeled every plot with the conf's first enabled signal type whenever `--first-channel`/`--channels` was passed explicitly to override the conf-derived range, regardless of which channels were actually requested (even though the correct `.dat` files were read, since `--file-prefix` is normally also given explicitly in that case).

## Test plan
- [x] `dll_pll_veml_tracking.cc` change builds cleanly against `next` (both GCC and Clang 20 toolchains)
- [x] `clang-format` (v22.1.8, matching CI) reports no changes needed
- [x] `clang-tidy-20` reports zero warnings for the touched lines (line-filtered to the diff)
- [x] Both Python files pass `ast.parse` / import cleanly; `SIGNAL_ORDER`/`SIGNAL_TYPES` consistency re-verified (15/15 entries)
- [x] Verified the channel-range label fix against multiple confs/ranges, including a range straddling two signal types, confirming per-channel (not per-run) resolution

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
+++ b/src/algorithms/tracking/gnuradio_blocks/dll_pll_veml_tracking.cc
@@ -1809,8 +1809,11 @@ void dll_pll_veml_tracking::log_data()
             uint64_t tmp_long_int;
             if (d_trk_parameters.track_pilot)
                 {
-                    prompt_I = d_Prompt_Data.data()->real();
-                    prompt_Q = d_Prompt_Data.data()->imag();
+                    // Same I/Q interchange as assign_correlators_to_synchro(): for signals whose
+                    // data and pilot components are true RF quadrature (e.g. Galileo E5a/E5b,
+                    // GPS L5), d_interchange_iq is true and the data bits are on the imaginary
+                    // axis of d_Prompt_Data, not the real one.
+                    map_correlator_to_iq(*d_Prompt_Data.data(), prompt_I, prompt_Q);
                 }
             else
                 {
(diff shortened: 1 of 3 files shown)


carlesfernandez:
Thanks! Merging

PR #1075, opened 2026-08-28
Fix Course Over Ground never being estimated

joebre:
## Summary
- `ground_speed_ms` in `Rtklib_Solver::get_PVT()` was declared and left at its `0.0` initializer instead of being assigned the computed ENU ground speed (`norm_rtk(enuv.data(), 2)`), which only ever reached `set_speed_over_ground()`.
- The `ground_speed_ms >= 1.0` gate that decides whether to compute Course Over Ground was therefore always false, so `cog` stayed permanently at its `-9999.0` "not estimated" sentinel regardless of actual receiver speed.
- Fix: assign the computed speed to `ground_speed_ms` before the threshold check, and reuse it for `set_speed_over_ground()`.

## Test plan
- [x] Rebuilt `pvt_libs` and the `gnss-sdr` executable cleanly with this change.
- [x] Verified `clang-format-19` (CI's pinned version) makes no changes to the touched file.


joebre:
Opened against the wrong base branch (main instead of next, per CONTRIBUTING.md). Superseded by #1076.

PR #1077, opened 2026-08-28
Add per-satellite azimuth/elevation and used signals to MonitorPvt

joebre:
## Summary
An application monitoring gnss-sdr (e.g. a sky plot, or a per-satellite
status view) typically needs some information about the satellites used
for the current PVT computation -- which satellites, which signal(s), and
their azimuth/elevation. Today `MonitorPvt` only exposes the scalar
solution (position/velocity/DOP/etc.), with no per-satellite breakdown.

This adds a `repeated UsedSatellite used_satellites = 37` field to
`MonitorPvt`, one entry per satellite/signal that contributed to the fix:

- `prn`, `system` (constellation letter: G/E/R/C/S/J)
- `signal` (2-char signal ID, e.g. "1C", "1B", "5X")
- `azimuth_deg`, `elevation_deg`
- `combined`: true when that satellite's signals were combined into the
  fix (e.g. Galileo E1+E5a iono-free combination) -- false for
  single-frequency

Signals are kept as separate entries rather than merged into one
per-satellite record: a satellite tracked on two signals that both
contribute to an iono-free combination appears as two entries, both
flagged `combined = true`, since a consumer may care which specific
signals were used, not just which satellites.

Populated in `Rtklib_Solver::get_PVT()` from RTKLIB's per-satellite
azimuth/elevation solution (`pvt_ssat[]`), intersected with the
observables map to find which signal(s) of each used satellite actually
contributed to that epoch.

## Test plan
- [x] Regenerated protobuf compiles (`monitor_pvt.pb.cc`) and `pvt_libs`
      builds cleanly against a fresh `next` checkout.
- [x] Verified `clang-format-22` (CI's pinned version on `next`) makes no
      changes to the touched files.
- [x] Field-tested against a live receiver in a companion monitor
      application: populated a per-channel azimuth/elevation table and a
      polar sky plot from this data across multiple cold/warm/hot-start
      cycles and combined/single-frequency configurations.

As an example of what this enables on the consuming side: here's a companion monitor application built on top of this data (per-channel Const./Sig./Az/El/PVT columns intersecting the channel table with `used_satellites`, a polar sky plot, and a "Used Satellites" count plot), running against a live Galileo E1B+E5X dual-frequency fix:
Screenshot from 2026-08-28 11-03-03

### DIFF
--- a/docs/protobuf/monitor_pvt.proto
+++ b/docs/protobuf/monitor_pvt.proto
@@ -52,4 +52,19 @@ message MonitorPvt {
 
   uint32 galhas_status = 35;  // Galileo HAS status: 1- HAS messages decoded and applied, 0 - HAS not available
   string geohash = 36;        // Encoded geographic location. See https://en.wikipedia.org/wiki/Geohash
+
+  // One satellite/signal used in this fix, with its azimuth/elevation and
+  // whether it was combined with another signal of the same satellite (e.g.
+  // Galileo E1+E5a iono-free combination). Signals are listed individually,
+  // not merged: a combined satellite appears as two entries, both with
+  // combined = true.
+  message UsedSatellite {
+    uint32 prn = 1;
+    string system = 2;  // "G" GPS, "E" Galileo, "R" GLONASS, "C" BeiDou, "S" SBAS, "J" QZSS
+    string signal = 3;
+    double azimuth_deg = 4;
+    double elevation_deg = 5;
+    bool combined = 6;
+  }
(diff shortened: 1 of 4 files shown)


carlesfernandez:
Wow, that's an awesome monitor tool! Thanks for your contributions.

vladisslav2011:
It got merged!!!
Maybe, it's time to add skyplot to gnss-sdr-monitor and stop using GPSD

PR #1083, opened 2026-09-04
Gate GPS L5/L2C and Galileo E5a/E5b assist acquisition on real broadcast status

joebre:
## Summary

`GNSSFlowgraph::search_next_signal()`'s dual-frequency assist path picks a
satellite for a secondary/assisted signal (GPS L5/L2C, Galileo E5a/E5b)
directly off its already-tracked primary frequency, bypassing normal
search entirely. It never checked whether that satellite's secondary
signal is actually usable, so:

- An older-block GPS satellite (no L5/L2C hardware) would be offered for
  L5/L2C acquisition over and over, forever.
- A Galileo satellite with a signal flagged unhealthy (independent of
  E1B) would be offered for that signal the same way.

Acquisition fails near-instantly in both cases (no real signal to find),
and the same fast path re-offers the identical satellite on the very next
idle tick, with no backoff. Observed live as a single `pcps_acquisition`
thread sustaining ~50-100% CPU indefinitely whenever the receiver's
longest-tracked primary-frequency satellite happened to lack the
secondary signal.

- **GPS**: decode the SV Configuration Code (IS-GPS-200, almanac page 25)
  into `Gps_Almanac::AS_status`, a field that already existed but was
  never populated -- the page 25 decode itself was marked with a literal
  `\TODO Read Anti-Spoofing, SV config`. Gate the assist pick on it:
  `code>=11` for L5, `code>=10` for L2C -- empirically validated against
  gnss-sdr's own PRN->block table (`Gnss_Satellite::what_block()`) for
  all 32 PRNs, zero exceptions where the table and the live broadcast
  agreed. Where they disagreed, the live broadcast was right: confirmed
  live with PRN 21 successfully tracking L5 despite the static table
  listing a pre-L5 block, since that table goes stale as satellites are
  replaced and the broadcast code doesn't.
- **Galileo**: gate on the already-decoded `E1B_HS`/`E5a_HS`/`E5b_HS`
  health flags (0 == OK per the ICD). Every current Galileo satellite
  broadcasts the full signal set by design, so there's no hardware gap to
  check here -- but a signal can still be flagged unhealthy independent
  of E1B. Also fixes a real, separate gap found along the way:
  `Galileo_Almanac_Helper::get_almanac()` case 2 decoded `E5b_HS` into
  `E5b_HS_9` but never copied it into the returned almanac, so any
  almanac-sourced (as opposed to ephemeris-sourced) health check for that
  slot always saw a stale default instead of the real value.

Both new checks stay permissive when no data exists yet for a satellite,
by design -- they only block on positive evidence, never on absence of
data.

## Test plan

- [x] Clean build from this branch in isolation (fresh `cmake` + build,
      no other local changes), zero errors/warnings.
- [x] Live-verified against a real dual-band GPS+Galileo capture:
  - GPS: a Block IIR-M satellite (no L5 hardware) was previously seen
    retrying L5 acquisition thousands of times over several minutes with
    a `pcps_acquisition` thread pegged near 100% CPU; with the fix, the
    same satellite is correctly never offered for L5, and CPU returns to
    baseline.
  - GPS: confirmed the SV Configuration Code decode itself against the
    known-correct block for all 32 PRNs currently in the constellation --
    codes group cleanly by generation (IIR, IIR-M, IIF all decode to a
    single consistent code each, with zero exceptions); the few PRNs
    where the live code disagreed with gnss-sdr's static table turned out
    to be cases where the *table* was stale, confirmed by that satellite
    successfully locking L5 live.
  - Galileo: with the fix in place, all tracked Galileo satellites
    correctly acquire their companion E5a signal (no regression -- no
    healthy satellite is ever blocked).

### DIFF
--- a/src/algorithms/PVT/adapters/rtklib_pvt.cc
+++ b/src/algorithms/PVT/adapters/rtklib_pvt.cc
@@ -926,6 +926,12 @@ std::map<int, Beidou_Dnav_Almanac> Rtklib_Pvt::get_beidou_dnav_almanac() const
 }
 
 
+std::map<int, Glonass_Gnav_Ephemeris> Rtklib_Pvt::get_glonass_gnav_ephemeris() const
+{
+    return pvt_->get_glonass_gnav_ephemeris_map();
+}
+
+
 void Rtklib_Pvt::connect(gr::top_block_sptr top_block)
 {
     if (top_block)
--- a/src/algorithms/PVT/adapters/rtklib_pvt.h
+++ b/src/algorithms/PVT/adapters/rtklib_pvt.h
@@ -229,6 +229,7 @@ public:
     std::map<int, Gps_Almanac> get_gps_almanac() const override;
     std::map<int, Galileo_Almanac> get_galileo_almanac() const override;
     std::map<int, Beidou_Dnav_Almanac> get_beidou_dnav_almanac() const override;
+    std::map<int, Glonass_Gnav_Ephemeris> get_glonass_gnav_ephemeris() const override;
 
     void connect(gr::top_block_sptr top_block) override;
     void disconnect(gr::top_block_sptr top_block) override;
--- a/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
+++ b/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
@@ -2103,6 +2103,12 @@ std::map<int, Beidou_Dnav_Almanac> rtklib_pvt_gs::get_beidou_dnav_almanac_map() 
 }
 
 
+std::map<int, Glonass_Gnav_Ephemeris> rtklib_pvt_gs::get_glonass_gnav_ephemeris_map() const
+{
+    return d_internal_pvt_solver->glonass_gnav_ephemeris_map;
+}
+
+
 void rtklib_pvt_gs::clear_ephemeris()
 {
     d_internal_pvt_solver->clear_gps_ephemerides();
--- a/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.h
+++ b/src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.h
@@ -57,6 +57,7 @@ class Galileo_HAS_data;
 class Galileo_Reduced_CED;
 class Geohash;
 class GeoJSON_Printer;
+class Glonass_Gnav_Ephemeris;
 class Gps_Almanac;
 class Gps_Ephemeris;
 class Gpx_Printer;
@@ -118,6 +119,11 @@ public:
      * \brief Get latest set of BeiDou DNAV almanac from PVT block
      */
     std::map<int, Beidou_Dnav_Almanac> get_beidou_dnav_almanac_map() const;
+
+    /*!
+     * \brief Get latest set of GLONASS GNAV ephemeris from PVT block
+     */
+    std::map<int, Glonass_Gnav_Ephemeris> get_glonass_gnav_ephemeris_map() const;
 
     /*!
      * \brief Clear all ephemeris information and the almanacs for GPS and Galileo
--- a/src/core/interfaces/pvt_interface.h
+++ b/src/core/interfaces/pvt_interface.h
@@ -27,6 +27,7 @@
 #include "beidou_dnav_ephemeris.h"
 #include "galileo_almanac.h"
 #include "galileo_ephemeris.h"
+#include "glonass_gnav_ephemeris.h"
 #include "gnss_block_interface.h"
 #include "gps_almanac.h"
 #include "gps_ephemeris.h"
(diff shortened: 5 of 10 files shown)


vladisslav2011:
Hello.
This PR looks good.
Up to me, a bit too much comments on obvious things and well-written self-documenting code, but this is not a problem, as these comments, I think, are intended to help your customers, who may be less experienced in GNSS signal processing/receiver design, to better understand what each line of the code does.

Questions:
1. What would happen if the L1 signal will be spoofed and replaced with a maliciously crafted signal, containing almanac with L2/L5 signal status set to unhealthy, but L2/L5 signal will be left intact? I think, the receiver will use spoofed signal and would never attempt to acquire the non-spoofed healthy L2/L5 band. Is it possible to handle this situation?
2. Is it possible to add the same secondary signal health check for other systems (QZSS, BeiDou, Glonass)?

Small suggestion: As almanac/ephemeris maps may be accessed from gnss_flowgraph at any arbitrary time now, the collision between map update event from telemetry decoder and map iteration in the gnss_flowgraph may happen with bad undesired effects... It may be better to guard access to these maps with a mutex (may be a subject to another "companion" PR).

joebre:
Hello, thanks for the feedback.

Done.
I don't have the impression that GNSS-SDR has been designed or structured to be spoofing-resistant. Spoofing resistance, if required, should in my view be enabled as an option rather than being the default behavior.

Reducing the Doppler uncertainty for the acquisition of a second signal is also not particularly spoofing-resistance-friendly. Moreover, as far as I remember, RTKLIB itself would not be resistant to a spoofing attack either.

If it helps address your concerns, I can make #1083 conditionally dependent on a configuration setting, enabled by default. This would allow users specifically targeting spoofing scenarios to disable this behavior and implement the required spoofing-resistance mechanisms in their system.

Almanac and Ephemeris concurrent access has been fixed in #1090

vladisslav2011:
In my experiments GNSS-SDR (with small modifications published here and some configuration option fine-tuning) was able to mitigate interference from chinese "anti-drone" device, that made all phones and some dedicated navigation devices fly in circles ~1000km away from the true location... This was not real spoofing, when false signals are slowly "pulling" the receiver away from true location. The device was just replaying the same simulated signal fragment over and over again and it had really bad LO stability.

> Almanac and Ephemeris concurrent access has been fixed in #1090

No. It is not fixed. Only LLM-generated comments mentioning the mutex usage got merged into next, not the actual `std::lock_guard<std::mutex> lck(d_almanac_ephemeris_map_mutex)`...

joebre:
Thanks for pushing back on this -- worth double-checking rather than taking my word for it.

I went back through `next` and traced the actual mechanism. It's not a mutex directly wrapping map access -- it's an immutable-snapshot swap:

- `rtklib_pvt_gs.h:272`: `mutable std::mutex d_snapshot_mutex;`
- `publish_navigation_snapshot()` builds a fresh `NavigationSnapshot` (each map copied into a new `shared_ptr<const ...Map>`), then swaps it in under the lock:
  ```cpp
  std::lock_guard<std::mutex> lock(d_snapshot_mutex);
  d_navigation_snapshot.swap(published);
  ```
- `navigation_snapshot()` (the read side) takes the same lock to copy out the current snapshot pointer:
  ```cpp
  std::lock_guard<std::mutex> lock(d_snapshot_mutex);
  return d_navigation_snapshot;
  ```
- Every `get_gps_ephemeris_map()` / `get_galileo_ephemeris_map()` / etc. -- what `PvtInterface::get_gps_ephemeris()` forwards to, which is what `gnss_flowgraph.cc`/`satellite_visibility.cc` actually call -- goes through `navigation_snapshot()`. `PvtInterface` never exposes the raw solver maps, so there's no path from the flowgraph/control-thread side that bypasses this.

So a reader never iterates a map that's concurrently being mutated: each publish is a copy-on-write, and the swap itself is what's locked, not the (already-immutable) map contents.

This has also been checked empirically, not just by code inspection: with a 2.5 h recording, run through `gnss-sdr-CtrlApp` with history enabled on the sky plot, any concurrent read/write corruption on the ephemeris maps should show up as visible spikes/outliers in the plotted satellite Az/El trajectories. Before this fix, those spikes were clearly visible. After it, no Az/El outliers have been observed over the same test.

`d_almanac_ephemeris_map_mutex` specifically doesn't appear anywhere in `next`'s history (checked with `git log --all -S`), so I can't find what that refers to -- possibly a name from an earlier draft or discussion rather than what actually got merged. If you're looking at a different snapshot of the code than current `next`, or I'm missing something, let me know which file/line and I'll take another look.

vladisslav2011:
>Almanac and Ephemeris concurrent access has been fixed in https://github.com/gnss-sdr/gnss-sdr/pull/1090

I have searched all commits from https://github.com/gnss-sdr/gnss-sdr/pull/1090 for `std::mutex` and `std::lock_guard` and have not found something named like `d_almanac_ephemeris_map_mutex` (this is just hypothesis, how it may be named, not the actual name in the code), so I thought, that it is not fixed, but it was fixed in a different PR.

joebre:
Added `GNSS-SDR.enable_secondary_signal_status_gating` (default `true`, so nothing changes unless you touch it) -- disabling it turns off all five broadcast-status checks added by this PR (GPS L5/L2C hardware-generation code, Galileo/BeiDou/GLONASS/QZSS secondary-signal health flags) and falls back to always attempting acquisition, same as before this PR. That gives anyone concerned about the spoofed-almanac scenario you described an easy opt-out without losing the benefit for everyone else.

PR #1091, opened 2026-09-07
Add an opt-in EVK1029_Signal_Source for the SAPHYRION EVK1029

joebre:
## Summary

EVK1029 is an Evaluation Kit for SAPHYRION SAGL space-grade ASICs: the SY1009 RF Front-End and the SY1019 ADC/DSP.

The kit includes:

- A USB 3.0 device connected to a dual-band (E1/E5a) or triple-band (E1/E5a/E6) GNSS front-end board.
- A Linux application for controlling and monitoring the EVK1029 USB device and the SY10x9 ASICs, and for saving the real-time captured GNSS data stream to a binary file on an HDD, SSD, or NVMe drive.
- An EVK1029-enabled version of GNSS-SDR.

EVK1029 is being developed under an ESA contract and is expected to become commercially available within 6-12 months.

This PR adds the GNSS-SDR side of that integration: `EVK1029_Signal_Source`, a signal source that reads the EVK1029 host application's raw capture files directly -- a continuous, header-less stream of OBA-encoded 4-bit samples (two per byte, 16 samples per little-endian 64-bit word), with no periodic block header. This format is distinct enough from the generic XML-metadata-driven `ION_GSMS_Signal_Source` path to warrant a dedicated reader rather than a configuration of it.

## Design notes

- **Multi-band sample lock.** `Evk1029Source` (the underlying GNU Radio block) supports emitting multiple identical output streams from a single read/file. A dual-band capture (e.g. two `Freq_Xlating_Fir_Filter` instances tuned to different IFs from the same raw stream) stays sample-locked between bands this way: GNU Radio computes `noutput_items` as the minimum available space across all declared output ports, so backpressure from whichever downstream chain is slower stalls production for every port together -- unlike two independently-reading `SignalSource` instances, which have no shared clock and can silently drift apart over a long run.
- **Scheduling quantum tied to sample rate.** The block's `output_multiple` (how much data the scheduler hands it per `work()` call) is derived from the configured `sampling_frequency` -- targeting ~0.7 ms of data, rounded to the nearest power of two -- rather than a hardcoded constant. This block runs at the raw, pre-decimation sample rate (up to ~180 Msps), where call-frequency overhead measurably affects both this block and every downstream reader woken at the same cadence; deriving the quantum from the actual rate keeps that tuning valid across different capture rates instead of only the one it happened to be tuned against.
- **Opt-in, disabled by default**, matching this project's convention for optional signal sources: build with `-DENABLE_EVK1029=ON` to enable it. A `CMakeLists.txt` option gates a compile definition on `core_receiver` and conditional registration in `gnss_block_factory.cc`.

## Configuration

```ini
SignalSource0.implementation = EVK1029_Signal_Source
```

|---|---|---|
| `filename` | `data.bin` | Path to the raw capture file produced by the EVK1029 host application. |
| `sampling_frequency` | falls back to `GNSS-SDR.internal_fs_sps` | The **raw, pre-decimation** ADC sample rate of the capture (not the decimated working rate downstream of `InputFilter`). Used to pace the optional throttle and to size the block's internal scheduling quantum (see above). |
| `RF_channels` | `1` | Number of identical, sample-locked output streams to emit from one underlying read -- set to N > 1 for a shared dual/triple-band capture feeding N downstream IF chains from a single file, instead of N independent `SignalSource` blocks each reading the file on their own. |
| `enable_throttle_control` | `false` | Paces output at real-time rate against `sampling_frequency`. Not supported with `RF_channels > 1` (a single throttle can't sit across multiple shared-source output ports); ignored with a warning in that case, since backpressure from the slower downstream chain already keeps every port in lockstep. |
| `dump` | `false` | Dump the raw unpacked byte stream to a file. |
| `dump_filename` | `./evk1029_signal_source.dat` | Where to write it, when `dump` is enabled. |

Sample format: real-valued (not complex), one signed byte per sample -- pair with `Pass_Through` or `Byte_To_Short` `DataTypeAdapter`s and a `Freq_Xlating_Fir_Filter` downstream, the same way other raw-byte-stream sources in this project are typically wired.

Example, dual-band Galileo E1+E5a / GPS L1+L5 capture sharing one file:

```ini
GNSS-SDR.num_sources = 1

SignalSource0.implementation      = EVK1029_Signal_Source
SignalSource0.filename            = /path/to/capture.bin
SignalSource0.sampling_frequency  = 178000000
SignalSource0.RF_channels         = 2
```

## Test plan

- [x] Clean build against `upstream/next` (this worktree) with the default configuration (`ENABLE_EVK1029` unset): zero EVK1029 code compiled in.
- [x] Clean build with `-DENABLE_EVK1029=ON`: `evk1029_signal_source.cc` and `evk1029_source.cc` compile and link into `gnss-sdr`.
- [x] Verified live against real EVK1029 captures (dual-band Galileo+GPS) over an extended local testing session: correct unpacking, correct sample-locked multi-band output, correct throttle/dump behavior.


carlesfernandez:
Thanks @joebre , please check if documentation at https://gnss-sdr.org/docs/sp-blocks/signal-source/#implementation-evk1029_signal_source is ok

PR #1092, opened 2026-09-07
Report az/el for satellites excluded from the PVT fix, not just used ones

joebre:
## Summary

`MonitorPvt.used_satellites` (added recently) reports one entry per
satellite/signal that contributed to the current fix, with its
azimuth/elevation. But a satellite that's tracked and has a valid
observation, yet gets excluded from the fix itself (below
`PVT.elevation_mask`, or dropped by RAIM FDE), got no entry at all --
identical, from the monitor's point of view, to a satellite with no
tracking/no data. That's misleading: it looks like a receiver problem
when it's actually the mask (or RAIM) working as intended.

## Root cause

`rescode()` (`rtklib_pntpos.cc`) computes each satellite's azimuth/elevation
via `satazel()` **before** checking it against `opt->elmin`
(`PVT.elevation_mask`), and `pntpos()` copies that `azel` into `ssat[]`
unconditionally -- only `ssat[].vs` is gated by the mask. The same is true
for a satellite excluded by RAIM FDE: `raim_fde()` only clears `vsat[i]`
for the excluded satellite, it doesn't touch its `azel` entry. So a valid
azimuth/elevation is already sitting in `ssat[]` for these satellites; the
`used_satellites` loop in `rtklib_solver.cc` was just discarding it by
gating the entire per-satellite entry on `vs`.

## Fix

- The `used_satellites` loop now keys on whether `azel` was actually
  computed this epoch (non-zero -- the same "no data" sentinel `pntpos()`
  itself resets `ssat[]` to at the top of every call), instead of `vs`.
- Added a `used` bool (default `true`) to `Monitor_Pvt::UsedSatelliteInfo`,
  the `UsedSatellite` protobuf message (field 7), and
  `serdes_monitor_pvt.h`'s read/write, reflecting `vs`.

A satellite that's tracked but excluded from the fix now still reports its
real azimuth/elevation, with `used = false`, instead of disappearing from
the list entirely.

## Test plan

- [x] Clean build against `upstream/next` (this worktree).
- [x] Full unit test suite (`run_tests`, `-DENABLE_UNIT_TESTING=ON`): 661
      passed, 1 skipped (unrelated `RtklibTlsTest`), 0 failed.
- [x] Verified live: added a temporary debug log of the raw `vs`/`azel`
      values RTKLIB computes for a satellite sitting a few degrees below
      `PVT.elevation_mask` -- confirmed `vs=0`, `used=0`, and a real,
      continuously-updating elevation value, matching the intended
      behavior; removed before this PR.

### DIFF
--- a/docs/protobuf/monitor_pvt.proto
+++ b/docs/protobuf/monitor_pvt.proto
@@ -61,7 +61,7 @@ message MonitorPvt {
   // azimuth_deg/elevation_deg populated with used = false when it was
   // tracked and had az/el computed but excluded from the fix itself (e.g.
   // below PVT.elevation_mask, or by RAIM FDE).
-  message UsedSatellite {
+  message TrackedSatellite {
     uint32 prn = 1;
     string system = 2;  // "G" GPS, "E" Galileo, "R" GLONASS, "C" BeiDou, "S" SBAS, "J" QZSS
     string signal = 3;
@@ -70,5 +70,5 @@ message MonitorPvt {
     bool combined = 6;
     bool used = 7;
   }
-  repeated UsedSatellite used_satellites = 37;
+  repeated TrackedSatellite tracked_satellites = 37;
 }
--- a/docs/protobuf/monitor_pvt.proto
+++ b/docs/protobuf/monitor_pvt.proto
@@ -53,11 +53,14 @@ message MonitorPvt {
   uint32 galhas_status = 35;  // Galileo HAS status: 1- HAS messages decoded and applied, 0 - HAS not available
   string geohash = 36;        // Encoded geographic location. See https://en.wikipedia.org/wiki/Geohash
 
-  // One satellite/signal used in this fix, with its azimuth/elevation and
-  // whether it was combined with another signal of the same satellite (e.g.
-  // Galileo E1+E5a iono-free combination). Signals are listed individually,
-  // not merged: a combined satellite appears as two entries, both with
-  // combined = true.
+  // One tracked satellite/signal with its azimuth/elevation, whether it was
+  // combined with another signal of the same satellite (e.g. Galileo E1+E5a
+  // iono-free combination), and whether it was actually used in this fix.
+  // Signals are listed individually, not merged: a combined satellite
+  // appears as two entries, both with combined = true. A satellite can have
+  // azimuth_deg/elevation_deg populated with used = false when it was
+  // tracked and had az/el computed but excluded from the fix itself (e.g.
+  // below PVT.elevation_mask, or by RAIM FDE).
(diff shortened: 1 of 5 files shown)


carlesfernandez:
It sounds a little odd to have a field named `used` in a message called `UsedSatellite`. Since this message was introduced fairly recently, do you think it would be too disruptive to rename it to `TrackedSatellite`, or perhaps simply `Satellite`?

joebre:
Good point -- renamed to `TrackedSatellite` (message + `tracked_satellites` field, and the matching C++ `Monitor_Pvt::TrackedSatelliteInfo`/`tracked_satellites`) in the latest commit. Not disruptive since the message is new enough that nothing depends on the old name yet.

vladisslav2011:
LGTM.
Up to me the possible improvements (ideas for this or another PR) may be the following:
 - Report tracked unhealthy satellites positions. This should be done carefully, not like I did it in my fork, as using unhealthy satellites for pseudorange calculation may have undesired effects on PVT accuracy.
 - Report untracked (due to obstackles in the signal path or insufficient channels), but visible satellites like hardware devices do. This may require merging https://github.com/gnss-sdr/gnss-sdr/pull/1090 or equivalent PR.

joebre:
I agree that Az/El computation for unhealthy satellites may produce inconsistent results, since there is no guarantee that the available ephemeris or almanac data are valid or usable.

This PR also relates to SV measurements eliminated by RAIM FDE.

However, if malformed or inconsistent ephemeris/almanac data can cause instability or unexpected behavior in the Az/El computation, I think this should be addressed directly at the Az/El computation level. The code there should be sufficiently fault-tolerant to safely handle invalid or inconsistent input data. This would prevent such issues upfront, regardless of which part of the application makes use of the computed Az/El values.

The same principle applies to pseudorange computation, which may include ionospheric corrections derived from potentially inconsistent navigation data. If there is evidence that invalid or inconsistent ionospheric parameters can cause problems, then, in my view, these cases should be handled safely at the ionospheric correction computation level rather than relying on higher-level code to prevent such data from reaching it.

Regarding "Report untracked", I fully agree. This is exactly what I would expect to see in a GNSS monitor application.

However, I don't want to address this before #1092 is merged into next. I need to keep the changes with respect to next as small as possible to get the PR approved and merged ASAP.

In my view, the proposed change should be implemented as an incremental improvement once both #1092 and #1090 have been merged.

I already have two other incremental PRs waiting for #1090 to be merged into next: runtime Doppler aiding from ephemeris and almanac data for visible satellites, and variable Doppler-grid PCPS/CFAR acquisition using [1...2N] Doppler bins, which unifies the various aided and unaided acquisition requirements.

PR #1093, opened 2026-09-08
Add BeiDou-3 B2a RNSS single-frequency SPP (B-CNAV2)

HuangChuhan-SYSU:
This PR adds a first-cut BeiDou-3 B2a (5D) RNSS single-frequency chain on top of `next`.

### What is included
- ICD BPSK(10) replica, PCPS acquisition, and DLL-PLL tracking on the 1 ms data component.
- B-CNAV2 decoder: preamble + CRC-24Q, MT10/MT11/MT30, systematic 288 bits.
- Channel ID `5D` / `BDS_B2A`. GEO and BDS-2 PRNs (1–18, 59–63) are not assigned B2a channels.
- B2a observations are paired only with CNAV2 (`sig_type = BDS_EPH_SOURCE_CNAV2`).
- B2a-only SPP remaps `5D` onto RTKLIB slot 0 with `lam[0]=c/FREQ5` (RTKLIB BDS `frq2` is B3, not L5).
- Sample file-input confs under `conf/File_input/Beidou/`.

### First-cut limitations 
- 64-ary LDPC is not decoded yet; the decoder uses the systematic 288 information bits + CRC-24Q.
- No FLL on live sky.

### Checks
- Commits are DCO-signed (`Signed-off-by: huangchuhan <huangchh37@mail2.sysu.edu.cn>`).
- clang-format 22 applied to the changed C/C++ files.
- Unit tests: `BeidouCnav2*:BeidouB2a*:BeidouB1cPvtHelpersTest.*` (28/28).
- LabSat File_130 B2a-only playback produced a single-frequency SPP fix.

I undertake the Contributor Covenant and the Developer’s Certificate of Origin.


src/algorithms/telemetry_decoder/gnuradio_blocks/beidou_b2a_telemetry_decoder_gs.cc
```diff
@@ -0,0 +1,324 @@
...
+beidou_b2a_telemetry_decoder_gs::~beidou_b2a_telemetry_decoder_gs()
+{
+    tlm_cleanup_and_save_files(d_dump_file, d_dump_filename, d_dump, d_dump_mat, d_remove_dat);
+}
+
+
+void beidou_b2a_telemetry_decoder_gs::set_satellite(const Gnss_Satellite& satellite)
+{
+    d_satellite = Gnss_Satellite(satellite.get_system(), satellite.get_PRN());
+    d_nav_msg_packet.prn = static_cast<int32_t>(d_satellite.get_PRN());
```
carlesfernandez, line 105: [P1] Clear accumulated navigation state when reassigning a channel

Changing the satellite here leaves `d_nav` intact, and `reset()` also leaves its accumulated ephemeris and `d_have_mt10/d_have_mt11/d_have_clk` flags intact. Consequently, after decoding MT10/MT11/MT30 for PRN 19, receiving just MT11 for PRN 21 makes the parser publish an ephemeris labelled PRN 21 containing retained PRN 19 orbit and clock fields. I reproduced this with CRC-valid synthetic frames using the actual parser. Channel::set_signal() reuses the decoder through this setter, so this can affect normal channel reassignment.

A localized fix for satellite reassignment is to reset the stream state and replace the parser when the satellite identity changes:

```suggestion
void beidou_b2a_telemetry_decoder_gs::set_satellite(const Gnss_Satellite& satellite)
{
    if (d_satellite.get_system() != satellite.get_system() ||
        d_satellite.get_PRN() != satellite.get_PRN())
        {
            reset();
            d_nav = Beidou_Cnav2_Navigation_Message{};
            d_TOW_at_current_symbol_ms = 0;
        }
    d_satellite = Gnss_Satellite(satellite.get_system(), satellite.get_PRN());
    d_nav_msg_packet.prn = static_cast<int32_t>(d_satellite.get_PRN());
```

This makes the parser require a complete new set after reassignment. For reset/reacquisition of the same satellite, also define whether accumulated navigation should be retained; if reset means starting a fresh decode, move the parser and TOW clearing into `reset()` itself.

Separately, in `decode_frame_symbols()`, reject a transmitted PRN mismatch after the CRC check and before setting `d_last_crc_ok` or calling `parse_info_bits()`:

```cpp
if (prn != 0U && frame_prn != prn)
    {
        return false;
    }
```

The existing `frame_prn` local already contains the transmitted PRN, so no additional parsing is needed.

src/core/system_parameters/beidou_cnav2_navigation_message.cc
```diff
@@ -0,0 +1,237 @@
...
+            parse_clock_common(bits, 42);
+        }
+    else if (mes_type == BEIDOU_CNAV2_MSG_CLK_DC)
+        {
+            parse_clock_common(bits, 64);
+        }
+
+    if (d_have_mt10 && d_have_mt11 && d_have_clk && d_eph.PRN < 59 && d_eph.sat_type != 1 && d_eph.hs == 0)
+        {
+            d_flag_new_eph = true;
```
carlesfernandez: [P1] Publish only a consistent ephemeris and clock set

These flags record whether each message type has ever arrived, rather than whether the accumulated fields belong to a consistent current data set. Once all three are true, every subsequent supported or unsupported CRC-valid message can cause publication. A new MT10 immediately replaces part of the orbit while retaining the old MT11 and clock fields. I reproduced immediate publication with a new toe of 6000 s while the previous clock retained toc = 3000 s; the issue is the absence of consistency validation, not a requirement that toe always equal toc. `parse_clock_common()` also sets `IODE = IODC`, so the independently broadcast ephemeris issue is not being checked.

Please decode the broadcast issue fields and assemble a candidate set using the B-CNAV2 issue/epoch consistency rules. Keep the last complete valid ephemeris separate from the candidate being assembled, invalidate incompatible candidate components on an issue change, and publish only when the required orbit and clock components are compatible. Do not manufacture IODE from IODC or republish merely because an unrelated message passed CRC.

src/algorithms/PVT/gnuradio_blocks/rtklib_pvt_gs.cc
```diff
@@ -2742,7 +2743,12 @@ int rtklib_pvt_gs::work(int noutput_items, gr_vector_const_void_star& input_item
                             if (!d_osnma_strict && tmp_eph_iter_bds_cnav1 != d_internal_pvt_solver->beidou_cnav1_ephemeris_map.cend())
                                 {
                                     const uint32_t prn_aux = tmp_eph_iter_bds_cnav1->second.PRN;
-                                    if ((prn_aux == gnss_synchro.PRN) && (std::string(gnss_synchro.Signal, 2) == std::string("1D")))
+                                    const int32_t eph_src = tmp_eph_iter_bds_cnav1->second.sig_type;
+                                    const std::string sig(gnss_synchro.Signal, 2);
+                                    // Same-PRN pairing: B1C↔CNAV1, B2a↔CNAV2.
+                                    if (prn_aux == gnss_synchro.PRN &&
+                                        ((sig == "1D" && eph_src == BDS_EPH_SOURCE_CNAV1) ||
+                                            (sig == "5D" && eph_src == BDS_EPH_SOURCE_CNAV2)))
```
carlesfernandez: [P2] Keep CNAV1 and CNAV2 ephemerides independently per PRN

The new source check correctly requires the matching navigation family, but both families still arrive as `Beidou_Cnav1_Ephemeris` and the message handler stores them in `beidou_cnav1_ephemeris_map[PRN]`. With B1C and B2a tracking the same satellite, each incoming family replaces the other. This condition then admits only whichever signal matches the latest stored record, so valid observations disappear according to navigation-message arrival order.

Please give CNAV2 its own ephemeris map or key the shared storage by both PRN and navigation family. Update ingestion, observable admission, solver lookup and output paths consistently. If simultaneous B1C/B2a operation is intentionally outside this first-cut scope, reject that configuration explicitly with a clear diagnostic rather than accepting it and silently overwriting navigation data.

src/algorithms/PVT/libs/rtklib_solver.cc
```diff
@@ -2609,6 +2643,10 @@ bool Rtklib_Solver::get_PVT(const std::map<int, Gnss_Synchro> &gnss_observables_
                         {
                             d_nav_data.lam[d_obs_data[k].sat - 1][0] = SPEED_OF_LIGHT_M_S / FREQ1;
                         }
+                    else if (is_bds_b2a_code(c0))
+                        {
+                            d_nav_data.lam[d_obs_data[k].sat - 1][0] = SPEED_OF_LIGHT_M_S / FREQ5;
+                        }
```
carlesfernandez: [P2] Set the B2a wavelength in the slot that contains its observation

This override checks only observation slot 0. The constructor remaps `5D` there only for `flags.check_only_enabled(BDS_B2A)`, which means B2a must be the only enabled signal across all constellations. For a GPS-L1 + B2a configuration, B2a remains in its default slot 2 and this override never runs. The preceding wavelength initialization assigns the BeiDou B3 wavelength to that slot, even though the observation is B2a. This gives the wrong carrier/Doppler scaling and broadcast-ionosphere frequency scaling.

Replace the slot-zero-only B2a branch with an independent loop over the measurement slots. This preserves the existing B1C handling and corrects B2a without changing the frequency indices shared with other constellations:

```suggestion
                    for (int band = 0; band < NFREQ; ++band)
                        {
                            if (is_bds_b2a_code(d_obs_data[k].code[band]))
                                {
                                    d_nav_data.lam[d_obs_data[k].sat - 1][band] = SPEED_OF_LIGHT_M_S / FREQ5;
                                }
                        }
```

The outer loop already excludes non-BeiDou satellites. This addresses the wavelength assignment; the downstream B2a slot-selection logic should also be audited for assumptions that B2a always occupies slot 0.

src/algorithms/telemetry_decoder/gnuradio_blocks/beidou_b2a_telemetry_decoder_gs.cc
```diff
@@ -0,0 +1,324 @@
...
+}
+
+
+void beidou_b2a_telemetry_decoder_gs::publish_navigation(double cn0_db_hz)
+{
+    if (d_nav.have_new_ephemeris())
+        {
+            auto eph = std::make_shared<Beidou_Cnav1_Ephemeris>(d_nav.get_ephemeris());
+            eph->PRN = d_satellite.get_PRN();
+            message_port_pub(pmt::mp("telemetry"), pmt::make_any(eph));
```
carlesfernandez, line 142: [P2] Route B-CNAV2 ephemerides to a matching RINEX navigation writer

Publishing this shared ephemeris type sends B-CNAV2 through the existing B-CNAV1 message handler and `log_rinex_nav_bds_cnav1()` path. For RINEX 4, that writer unconditionally emits a `CNV1` record header (rinex_printer.cc:3952) and writes `ISC_B1Cd` rather than `ISC_B2ad` (line 4026). Therefore B2a navigation output is labelled as B1C and loses its data-component delay correction. Adding the observation code alone does not make the navigation output correct.

Please dispatch navigation serialization by the ephemeris source and implement the corresponding CNV2 record layout and delay fields. Update both initial navigation-file output and subsequent ephemeris updates, including any deduplication that currently compares only PRN/toe. If native CNV2 output is deferred, explicitly suppress unsupported records and document the limitation rather than writing them as CNV1.

carlesfernandez (changes requested):
I recommend requesting changes before merging this first-cut B2a implementation. The inline comments identify two navigation-state correctness problems and three integration problems involving ephemeris storage, observation wavelengths and RINEX output. The documented omission of LDPC decoding is not itself the basis for this recommendation.

Validation performed on commit `849c18b0e64e111cf851dd4c403ca3c6a7853636`:
- Compiled and ran the 28 focused B-CNAV2, B2a replica/PVT-helper and existing B1C PVT-helper tests in a standalone test executable: all passed.
- Ran additional CRC-valid synthetic-frame reproductions against the actual parser: it published mixed satellite data after a PRN change, published partially replaced navigation data, and accepted a frame whose transmitted PRN differed from the requested channel PRN.
- The full local run_tests build stopped in existing generated GPS/Galileo Protobuf sources, outside the PR changes. This is a validation limitation, not a build regression attributed to this PR.
- No live-signal playback was performed.

Please address the inline findings before merging. Small, localized fixes are included as code suggestions; the navigation assembly, storage and RINEX changes require coordinated updates.

PR #1095, opened 2026-09-10
PCPS acquisition second step alignment experiment

vladisslav2011:
This PR adds code boundary alignment to PCPS acquisition second step in two step configuration.
This improves Doppler shift estimation accuracy when a bit transition is located close to the middle of the acquisition buffer and significantly reduces probability of false lock in the tracking block.
Improved acquisition performance of following signals:
 - Galileo E1b
 - BeiDou B1C

New configuration option `Acquisition_XX.aligned_step2`, defaults to true. Set to false to disable alignment for testing purposes.

Recommended acquisition configuration (Galileo E1b):
```
Acquisition_1B.implementation=Galileo_E1_PCPS_Ambiguous_Acquisition
Acquisition_1B.pfa=0.025
Acquisition_1B.doppler_max=4600
Acquisition_1B.doppler_step=100
Acquisition_1B.max_dwells=4
Acquisition_1B.second_doppler_step=31
Acquisition_1B.second_nbins=9
Acquisition_1B.make_two_steps=true
```

Recommended acquisition configuration (BeiDou B1C):
```
Acquisition_1D.implementation=BEIDOU_B1C_PCPS_Ambiguous_Acquisition
Acquisition_1D.item_type=gr_complex
Acquisition_1D.acquire_pilot=true
Acquisition_1D.max_dwells=4
Acquisition_1D.qmboc=true
Acquisition_1D.pfa=0.05
Acquisition_1D.doppler_max=4600
Acquisition_1D.doppler_step=100
Acquisition_1D.second_doppler_step=27
Acquisition_1D.second_nbins=13
Acquisition_1D.make_two_steps=true
```

Original idea by @joebre (see https://github.com/gnss-sdr/gnss-sdr/pull/1071)

 - [x] Field testing: Galileo E1b
 - [x] Field testing: BeiDou B1C
 - [x] Field testing: Interaction with non-blocking mode - no issues
 - [x] Field testing: Interaction with `bit_transition_flag` - no issues
 - [x] Field testing: Interaction with `coherent_integration_time_ms`=4 (Galileo) - no issues
 - [x] Field testing: Interaction with `GNSS-SDR.use_acquisition_resampler`=true - no issues
 - [x] Regression tests
 - [x] Examples update

Recommended addition: https://github.com/gnss-sdr/gnss-sdr/pull/1096


joebre:
I'm still not convinced that this is the right place to move my modifications. Conceptually, what we're doing is more part of the pull-in phase than the acquisition phase (but maybe I'm "biased", since in all the receivers I've worked on, I implemented this as part of tracking).

Anyway, the important thing is to get this feature merged into next. If that requires moving it into another PR, that's fine with me.

Please don't forget to also port Acquisition_1B.full_grid_search, which is required to eliminate or avoid the remaining unavoidable loss-of-lock cases for strong signals. 

**Question**: does Galileo_E1_DLL_PLL_VEML_Tracking is safe against secondary code peaks locking ? I got some cases in which Galileo E1 locks on perfectly stable tracking point (0.55 chips from the main peak with  early_late_space_chips = 0.15 and very_early_late_space_narrow_chips = 0.3) with a Cn0 being considerably below the one I get in other runs (Doppler correct and bit sync).

vladisslav2011:
Up to me, this is not attempt to move your modifications to a different place. This is completely different modification to existing 2-step acquisition implementation, that is based on ideas from discussions on your PR. I thought, that similar result may be achieved with significantly less code changes, tried to implement it, proved, that it performs well and opened this PR. I'm not sure that this is the only possible solution for "bit transition: problem as there is `bit_transition_flag` parameter, that enables another part of code attempting to solve this problem slightly differently (without much success).
`Acquisition_1B.full_grid_search` may be included here or it may be a subject for a different PR. Galileo still shows some "Loss of lock" events, one or two per 10 runs, BieDou shows less than 1 false lock per 10 runs.
Maybe I'll add `Acquisition_1B.full_grid_search` later.
I don't know will this be merged into `next` or not. Both may get rejected or both may get merged.
This PR is made the way, it can be applied independent of #1071.

>Question: does Galileo_E1_DLL_PLL_VEML_Tracking is safe against secondary code peaks locking ?

It depends on tracking block settings. The tracking block may be configured the way, it will lock onto a false correlation peak. This may happen during transition to narrow tracking as transition is performed as a momentary change of some tracking parameters. If narrow tracking loops bandwidths are significantly lower, than initial tracking loops bandwidths bad things usually happen. There is draft PR, that tries to avoid loss of lock and jumping to a false correlation peak during transition: https://github.com/gnss-sdr/gnss-sdr/pull/1049, but it has some issues itself. For example, 3-rd order DLL filter becomes unstable during transition/at very low bandwidth, so either transition should be disabled or a warning about suboptimal configuration should be emitted if the 3-rd order DLL filter is enabled... And I have no idea, how to test it. Maybe, updating default position test with a new configuration will be enough.

PR #1042, opened 2026-07-08
acquisition: take into account detected clock drift and implement doppler search grid size reduction

vladisslav2011:
As RTKLIB gives us a clock drift estimation, we can take it into account when doing acquisition if we have a valid PVT.

Added new configuration options:
wide_doppler_max - wide doppler search range used when no clock drift estimation and no assistance is available. Defaults to doppler_max, but may be increased to take into account frontend clock inaccuracy.
wide_doppler_step - wide doppler search step used when no clock drift estimation and no assistance is available. Defaults to doppler_step, but may be increased to reduce memory footprint/speedup initial acquisition.

narrow_doppler_max - narrow doppler search range used when assistance from a different signal from the same satellite is available. Defaults to 50 Hz.
narrow_doppler_step - narrow doppler search step used when assistance from a different signal from the same satellite is available. Defaults to 50 Hz.

This should reduce overall CPU consumption from acquisition during long runs and may improve acquisition performance.

`set_doppler_center` method is renamed to `set_assistance` as it updates both center frequency and range/step preset at once now.

- [x] Add `double get_clock_drift_ppm()` method to PVT interface/block/adapter
- [x] Update acquisition manager to adjust doppler grid center frequency using detected clock drift
- [x] Add `void set_assistance(int /*doppler_center*/, int32_t /*assist level*/)` to acquisition interface
- [x] Update acquisition manager to pass correct assist level to acquisition implementation
- [x] Update `pcps_acquisition`
- - [x] Write minimal tests
- - [x] Update examples
- - [x] Update documentation
- [x] Update `Galileo_E5a_Noncoherent_IQ_Acquisition_CAF`
- - [x] Write minimal tests
- - [x] Update examples
- [ ] Update `pcps_acquisition_fine_doppler`
- - [ ] Write minimal tests
- - [ ] Update examples
- [ ] Update `pcps_tong_acquisition`
- - [ ] Write minimal tests
- - [ ] Update examples
- [ ] Update `pcps_quicksync_acquisition`
- - [ ] Write minimal tests
- - [ ] Update examples
- [ ] Update `pcps_cccwsr_acquisition`
- - [ ] Write minimal tests
- - [ ] Update examples
- [ ] Update `galileo_pcps_8ms_acquisition`
- - [ ] Write minimal tests
- - [ ] Update examples
- [ ] Update `pcps_opencl_acquisition`
- - [ ] Write minimal tests
- - [ ] Update examples
- [ ] Update `pcps_acquisition_fpga`
[...]

PR #1097, opened 2026-09-11
Unify pcps_acquisition's Doppler search into a single N-bin API

joebre:
## Summary

`set_doppler_uncertainty()` only supported two states (full grid, or a single exactly-known bin, added in #1078). Replaces it with `set_doppler_num_bins(uint32_t)`: `0` = full configured grid (sentinel, since that count is otherwise private to the class), any other value = literal candidate bin count. Regular and narrowed searches now share one code path instead of two.

- CFAR reference-row placement generalized via `needs_extra_reference_row(candidate_bins)`, covering both the full-grid and narrowed cases with one formula. New `reference_bin_min_sidelobes` config (default `4`) drives it automatically.
- Fixes a latent asymmetry: with more than one real candidate, two reference rows are now computed (`+doppler_max`/`-doppler_max`), and the statistic picks whichever sits opposite the winning candidate's side of center, instead of always using a fixed `+doppler_max` row that could end up next to the signal.
- Detection threshold recalibrates whenever the active candidate count changes.
- Rename propagated through `AcquisitionInterface`, `PcpsAcquisitionAdapter`, `ChannelInterface`, `Channel`, `GNSSFlowgraph`'s two call sites.

## Behavior change

`next`'s `Acquisition_<Sig>.enable_doppler_narrowing` (#1078, default `false`) is removed, not renamed. Today, leaving it at default means an assisted acquisition still searches the full grid. After this PR, an assisted request (`N == 1`) always narrows to exactly 1 bin, unconditionally -- the caller's `N` is authoritative everywhere, no flag can override it. If you use `assist_dual_frequency_acq` without `enable_doppler_narrowing = true` today, this changes your receiver's behavior. Any `.conf` still setting that key just has it silently ignored.

A caller wanting a safety margin instead of full trust in its assist estimate should request more than 1 bin directly, not rely on a flag changing what "1" means.

## Configuration

|---|---|---|
| `Acquisition_<Sig>.reference_bin_min_sidelobes` | `4` | Target sidelobe clearance (in Doppler) for deciding whether a dedicated CFAR reference row is needed -- never used to place it beyond `doppler_max`. Only takes effect with CFAR (`pfa > 0`). |

## Known limitation

Searching only a few Doppler bins -- which assisted acquisition now always does -- isn't statistically free: `compute_threshold()`'s PFA calibration and the CFAR noise-floor estimate both rely on averaging over fewer independent hypotheses at small candidate counts than at the full grid, so detection is less statistically robust there. Pre-existing property of the design, not introduced by this PR -- flagged since assisted acquisition now always runs at that small a count, not just when a flag was opted into.

## Test plan

- [x] Full local test suite (`acq_test`, `trk_test`, `gnss_block_test`, `flowgraph_test` -- 105 tests) passes, from-scratch build off `upstream/next`.
- [x] `clang-format` clean against the CI-pinned version (22).

### DIFF
--- a/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.cc
+++ b/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.cc
@@ -303,9 +303,9 @@ void PcpsAcquisitionAdapter::set_doppler_center(int doppler_center)
 }
 
 
-void PcpsAcquisitionAdapter::set_doppler_uncertainty(unsigned int doppler_uncertainty)
+void PcpsAcquisitionAdapter::set_doppler_num_bins(unsigned int doppler_num_bins)
 {
-    acquisition_->set_doppler_uncertainty(doppler_uncertainty);
+    acquisition_->set_doppler_num_bins(doppler_num_bins);
 }
 
 
--- a/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.h
+++ b/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.h
@@ -103,9 +103,10 @@ public:
     void set_doppler_center(int doppler_center) override;
 
     /*!
-     * \brief Set Doppler uncertainty for the grid search
+     * \brief Set the number of Doppler bins to search, centered on
+     * set_doppler_center() -- see pcps_acquisition::set_doppler_num_bins().
      */
-    void set_doppler_uncertainty(unsigned int doppler_uncertainty) override;
+    void set_doppler_num_bins(unsigned int doppler_num_bins) override;
 
     /*!
      * \brief Returns the maximum peak of grid search
--- a/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
+++ b/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
@@ -131,11 +131,18 @@ pcps_acquisition::pcps_acquisition(const Acq_Conf& conf_)
       d_magnitude_grid_stride(aligned_row_stride<float>(d_effective_fft_size)),
       d_doppler_wipeoffs_stride(aligned_row_stride<gr_complex>(d_fft_size)),
       d_num_doppler_bins(static_cast<uint32_t>(std::ceil(static_cast<double>(2 * d_doppler_max) / static_cast<double>(d_doppler_step)))),
-      d_num_doppler_bins_step1_capacity(std::max(d_num_doppler_bins, 2U)),
       d_num_doppler_bins_step2(conf_.num_doppler_bins_step2),
       d_dump_channel(conf_.dump_channel),
-      d_threshold(conf_.pfa > 0.0 ? compute_threshold(conf_.pfa, d_effective_fft_size, d_num_doppler_bins, conf_.bit_transition_flag ? 1 : conf_.max_dwells) : conf_.threshold),
-      d_threshold_narrowed(conf_.pfa > 0.0 ? compute_threshold(conf_.pfa, d_effective_fft_size, 1U, conf_.bit_transition_flag ? 1 : conf_.max_dwells) : conf_.threshold),
+      d_min_reference_separation_hz((static_cast<float>(conf_.reference_bin_min_sidelobes) + 0.5F) / (static_cast<float>(conf_.sampled_ms) / 1000.0F)),
+      // Inline, not via needs_extra_reference_row(candidate_bins) -- that
+      // helper reads d_use_CFAR_algorithm_flag, declared (so initialized)
+      // later in this same list; see its own doc comment in the header.
+      d_full_grid_reference_needs_extra_row(conf_.use_CFAR_algorithm_flag && (static_cast<float>(d_num_doppler_bins / 2) * static_cast<float>(d_doppler_step) < d_min_reference_separation_hz)),
+      // A full grid worth needing a dedicated reference row for is always
+      // > 1 candidate bin in practice (see d_num_reference_rows_active's doc
+      // comment for why the count differs at exactly 1 candidate) -- so this
+      // ceiling reserves 2 rows, not 1, whenever the row is needed at all.
+      d_num_doppler_bins_full_grid_active(d_num_doppler_bins + (d_full_grid_reference_needs_extra_row ? (d_num_doppler_bins > 1U ? 2U : 1U) : 0U)),
       d_threshold_step_two(conf_.pfa2 > 0.0 ? compute_threshold(conf_.pfa2, d_effective_fft_size, d_num_doppler_bins_step2, conf_.bit_transition_flag ? 1 : conf_.max_dwells) : conf_.threshold),
       d_cshort(conf_.it_size != sizeof(gr_complex)),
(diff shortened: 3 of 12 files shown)


vladisslav2011:
Hello.
I don't think, that dedicated reference row is required for CFAR. The same row contains enough uncorrelated noise, so excluding the detected peak and it's neighborhood should be enough to provide a good reference for CFAR.
\+ small hint: you may enable the GitHub CI in your repository and eliminate the need to push empty commits to just re-trigger the CI in the PR. Re-triggering only failed jobs may be a lot faster, than re-running all jobs including heavy clang-tidy job.

joebre:
Good point on the full-grid case -- that's actually already the default behavior: `needs_extra_reference_row()` reuses the in-grid "opposite bin" (the existing candidate span minus the peak) as the noise reference, and only computes a dedicated extra row when CFAR is active *and* that natural distance can't clear `reference_bin_min_sidelobes`. So for a full grid, your suggestion is what happens unless the grid is too small to clear the sidelobe-clearance target on its own.

Where it doesn't work the same way is the narrowed/assisted case this PR is actually for -- as few as **1** candidate bin (an exactly-known Doppler from dual-frequency assist). The reason for the dedicated row there isn't just "no spare bins" -- it's about what CFAR is actually promising: `Acquisition_<Sig>.pfa` sets a *target false-alarm rate*, and CFAR is supposed to deliver that rate specifically by calibrating the threshold against a genuinely noise-only reference. With a single (or strongly reduced) bin count, it's essentially unavoidable that any in-row "reference" is itself influenced by the signal or its sidelobes present in that same correlation vector -- there's no bin left that's clearly signal-free. That doesn't stop detection (the signal still triggers), but it means the resulting `pfa` is no longer the one actually configured -- the statistical guarantee breaks even though the mechanism still "works."

So this is really a question of what you want: if an accurate `pfa` in the narrowed case matters, the dedicated reference row is what preserves it. If you don't care about a correct `pfa` there and detection alone is enough, I can switch to the opposite mechanism (reusing in-row bins the way the full-grid case does) for the `candidate_bins == 1` case specifically. Let me know which you'd prefer.

vladisslav2011:
I prefer something like this: https://github.com/gnss-sdr/gnss-sdr/pull/1042
3 different sets of presets for `doppler_max`/`doppler_step` covering 3 special cases:
 - Initial wide search with `wide_doppler_max` including Doppler shift from satellite motion, receiver motion and some room for receiver clock error. In case of HackRF without TCXO/ADLM Pluto it may be really high, 20...40ppm or even more, so `wide_doppler_max` may be 65000 Hz.
 - First narrowing stage when a PVT with minimum number of satellites is obtained. `doppler_max` may be reduced to ~5000Hz (3000Hz at L2/L5), `doppler_step` may be reduced too. 
 - Second narrowing stage when some sort of assistance is available  (other band signal with good telemetry/almanac/ephemeris) and a PVT is obtained : 1 or 2 bins, depending on `narrow_doppler_max`/`narrow_doppler_step` values
 - Maybe a different 

So, up to me, it may be better to make everything configurable.

Offtop:
Does your hardware support realtime input configurations?
If so, Are there any plans on implementing realtime input from EVK1029?
What's your position on realtime configuration performance optimizations (Pass_Through with set_output_multiple, that has no effect on non-throttled file input configurations, but has significant effect on realtime input configurations, from full inability to track anything to fully stable several days runs, or, for example, something like this https://github.com/gnss-sdr/gnss-sdr/pull/1108 that makes processing large bandwidths possible in realtime input configurations)?

joebre:
Hello, as I mentioned in #1042, I think it would make more sense to have the LO accuracy (in ppm) and max_receiver_dynamic as configuration parameters, while keeping Doppler_max as the pure geometric contribution to the Doppler.

This would allow the system to compute the appropriate Doppler uncertainty on its own, according to its current "state of knowledge", without the need to define special cases. It would also help the end user prepare an optimal .conf, since these parameters describe platform/receiver characteristics that are typically well known to the user.

The idea behind #1097 is to make pcps_acquisition more flexible, so that it can properly handle any required Doppler uncertainty, rather than defining how it should be used in specific scenarios.

Regarding the real-time aspect: in practice, our FE supports real-time data transfer to the host with some margin for both the 2- and 3-band versions (175 MHz / 4-bit and 215 MHz / 4-bit). Considering the performance I now get from GNSS-SDR — dual-band processing is faster than real time — we could potentially run GNSS-SDR directly in real time.

Real-time operation was not originally planned, mainly because of the performance we observed in our early GNSS-SDR tests. However, given the current results, if I have some spare time I may investigate this possibility as well.
What is your opinion on the most promising approach for this? Would it be better to implement/use a native GNSS-SDR source, e.g. TCP/UDP/FIFO, or would implementing the source as a GNU Radio block be the cleaner and more maintainable solution?

vladisslav2011:
> Would it be better to implement/use a native GNSS-SDR source, e.g. TCP/UDP/FIFO, or would implementing the source as a GNU Radio block be the cleaner and more maintainable solution?

Your existing `Evk1029Source` seems to support reading a FIFO, so adding an example configuration with a comment describing this approach may be enough. On the other hand, adding a full featured GNU Radio block with ability to configure the hardware may be more interesting, but more time consuming and harder to maintain in the future...

carlesfernandez:
Hi @joebre,

I want to merge this change (and #1106) before the release of GNSS-SDR v0.0.22. Working on it.

carlesfernandez:
I would lean toward implementing the source as a GNU Radio block, especially if you want this to become a reusable real-time source rather than something specific to one GNSS-SDR configuration. That would keep the data source naturally integrated into the GNU Radio flowgraph and, at the same time, make it potentially useful outside GNSS-SDR as well, broadening the ecosystem of tools that can interact with the hardware.

That said, I agree with @vladisslav2011 that we should avoid putting too much hardware-specific configuration logic into the block unless there is a clear benefit. If the existing EVK1029 software can already expose the sample stream through a FIFO, TCP, or UDP interface, reusing `Evk1029Source` (or another generic GNU Radio source) would probably be the simplest and most maintainable first approach.

If later you find that tighter integration is useful (for example, configuring sampling rate, bands, gains, starting/stopping streaming, etc. directly from the GNSS-SDR configuration) then a dedicated GNU Radio source block would make more sense.

joebre:
@carlesfernandez: hello, will be back on gnss-sdr by tomorrow...

joebre:
Hi @charlesfernandez,

Thanks for the update — we'll put #1097 and #1106 on standby and hold off on any further changes while you work on merging them for v0.0.22.

One bit of context that might help while resolving the conflict with `next`: #1097 is meant to fully replace the single-bin special case (`d_doppler_search_narrowed` / `set_doppler_uncertainty()`) that an earlier PR introduced and that's currently in `next`. That special case was written specifically to collapse the Doppler search to a single bin when a primary signal's already-tracked Doppler assists a secondary signal's acquisition. #1097 generalizes that into one N-bin API (`set_doppler_num_bins()`), so the assisted single-bin case, a narrowed window, and the full grid all go through the same code path instead of a separate branch — the single-bin case is just N=1 of the general mechanism.

We're open to any changes or restructuring on your end that make the integration easier — just let us know if there's anything we can do to help.

PR #1101, opened 2026-09-14
Optimize acquisition sample copy

MathieuFavreau:
Goal of this PR is to reduce unnecessary work while filling the sample buffer for acquisition:
- Instead of copying the samples to `d_data_buffer` and then to `d_input_signal`, just copy directly to `d_input_signal`
- Pre pad the `d_input_signal` buffer with zeros when `d_fft_size` is bigger than the number of samples to consume
- Get to state 2 faster (because of check order, it used to call `general_work` to copy 0 samples, and then go to state 2)

In addition to these changes, I've renamed some variables to make it more obvious what they represent

Edit: Optimized cshort as mentioned in [vladisslav2011](https://github.com/vladisslav2011) comment


vladisslav2011:
Hello.
This PR looks good.
Up to me, it may be better to split refactoring to a separate commit, but it may be merged as is.
Some thoughts on PCPS acquisition optimization:
 - Remove `d_data_buffer_sc` and call `volk_gnsssdr_16ic_convert_32fc` from `general_work`
 - Add `forecast` method to notify GNU Radio scheduler about the number of samples, that we may want to see in the input buffer and eliminate multiple `std::copy` calls in `general_work`. The number of consumed samples may need to be adjusted in this case to reduce negative side effects on acquisition performance...
 - Add optional "optimal grid" mode. In this mode adjust the Doppler step to nearest T/N and reduce the number of forward FFTs and number of Doppler wipeoffs to minimum required (N) as other Doppler wipeoffs may be obtained by circular shifting the FFT output several bins left/right. This should significantly improve the acquisition performance, up to ~2 times faster for signals with long code (GPS L2C/Galile E1b/BeiDou B1C)
 - Make `blocking=false` mode work as expected. I've almost fixed this, so maybe, Ill open an issue and a corresponding PR...
 
Upd: fixed some typos

MathieuFavreau:
> Hello. This PR looks good. Up to me, it may be better to split refactoring to a separate commit, but it may be merged as is. Some thoughts on PCPS acquisition optimization:
> 
> * Remove `d_data_buffer_sc` and call `volk_gnsssdr_16ic_convert_32fc` from `general_work`
> * Add `forecast` method to notify GNU Radio scheduler about the number of samples, that we may want to see in the input buffer and eliminate multiple `std::copy` calls in `general_work`. The number of consumed samples may need to be adjusted in this case to reduce negative side effects on acquisition performance...
> * Add optional "optimal grid" mode. In this mode adjust the Doppler step to nearest T/N and reduce the number of forward FFTs and number of Doppler wipeoffs to minimum required (N) as other Doppler wipeoffs may be obtained by circular shifting the FFT output several bins left/right. This should significantly improve the acquisition performance, up to ~2 times faster for signals with long code (GPS L2C/Galile E1b/BeiDou B1C)
> * Make `blocking=false` mode work as expected. I've almost fixed this, so maybe, Ill open an issue and a corresponding PR...
> 
> Upd: fixed some typos

I've made the change for cshort you mentioned, the other changes are outside the scope of this PR.
I agree that separating refactor from changes is a good practice, but I felt since I only renamed 3 variables directly linked to the changes, it wasn't too much a burden to review and understand where the logic change are, but I can split it if requested

For the `blocking=false` issue are you referring to issues with non blocking and 2 step acquisition?

vladisslav2011:
> the other changes are outside the scope of this PR.

These are just my ideas for a different PR. If you have plans to implement some of these changes, you may notify me just to avoid creating different PRs, that will do the same thing slightly differently with different configuration parameters and so on.

> are you referring to issues with non blocking and 2 step acquisition?

No. This is completely different issue https://github.com/gnss-sdr/gnss-sdr/issues/1102. The fix may include different implementation of https://github.com/gnss-sdr/gnss-sdr/pull/1014, but only in non-blocking mode (as it would be easy to perform the second step on already buffered samples). It may include re-implementation of https://github.com/gnss-sdr/gnss-sdr/pull/1095 as well (with one less non-coherent integration during second step), but again, only in non-blocking mode.

MathieuFavreau:
> > the other changes are outside the scope of this PR.
> 
> These are just my ideas for a different PR. If you have plans to implement some of these changes, you may notify me just to avoid creating different PRs, that will do the same thing slightly differently with different configuration parameters and so on.
> 
> > are you referring to issues with non blocking and 2 step acquisition?
> 
> No. This is completely different issue #1102. The fix may include different implementation of #1014, but only in non-blocking mode (as it would be easy to perform the second step on already buffered samples). It may include re-implementation of #1095 as well (with one less non-coherent integration during second step), but again, only in non-blocking mode.

Ah yes `max_dwells != 1` is causing blocking behavior. 

I was planning on opening a new PR similar to #1014 but as a config options instead, because I feel like the performance change might not be acceptable from the lack of responses I had on that PR.

vladisslav2011:
> I was planning on opening a new PR similar to #1014

Up to me, adding a commit with configuration option implementation and rebasing the PR on top of current `next` may be enough.

> the lack of responses I had on that PR

There was a top-priority PR https://github.com/gnss-sdr/gnss-sdr/pull/1093 adding a new really interesting signal. Adding a signal, that gives 15 seconds cold start to first fix time and ~9 seconds hot start to first fix time (at L5 band!) seems significantly more important than various small acquisition tweaks...
I'll take a look at #1014 and leave some feedback.

MathieuFavreau:
> > I was planning on opening a new PR similar to #1014
> 
> Up to me, adding a commit with configuration option implementation and rebasing the PR on top of current `next` may be enough.
> 
> > the lack of responses I had on that PR
> 
> There was a top-priority PR #1093 adding a new really interesting signal. Adding a signal, that gives 15 seconds cold start to first fix time and ~9 seconds hot start to first fix time (at L5 band!) seems significantly more important than various small acquisition tweaks... I'll take a look at #1014 and leave some feedback.

That's fine, I understand this is an open source project and bandwidth is limited, I was just mentioning because there was no comments in 4 months of the PR being open

carlesfernandez:
@MathieuFavreau My bad. I'm struggling to keep up with PR reviews while also trying to maintain an "always usable" `next` branch. I really appreciate your contributions, which is precisely why I tend to spend quite a bit of time reviewing them carefully :-)

MathieuFavreau:
@carlesfernandez Don't worry about it, very grateful for this repository. I was just mentioning that perhaps I needed to change the PR #1014 to make it easier to integrate with lower risk. I know I've introduced a couple of bugs in `next` in the past with my refactors, so your concern is justified haha

PR #1103, opened 2026-09-15
Add a per-PRN-and-signal acquisition retry cooldown

joebre:
## Summary

With few (or one) entries in a signal's visible bucket, a satellite that fails acquisition immediately (too weak to lock yet, right at the elevation mask, etc.) gets requeued by `apply_action()` case 0 and can be handed straight back out to the next idle channel on the very next idle tick. With enough idle channels polling `acquisition_manager()` at ~10 Hz, the same PRN can be re-attempted hundreds of times per second for no benefit -- observed live as sustained CPU usage with zero progress once a signal's search pool narrows to a single candidate.

`GNSS-SDR.acquisition_max_retry_rate_hz` (`0` = disabled, preserving prior behavior) caps how often any single PRN+signal can be handed out for a fresh acquisition attempt. Applied uniformly across every path that starts one: `pop_by_visibility()`, the legacy (non-visibility-aware) pick path, the primary-frequency dual-band assist fast path, and `apply_action()` case 2 (TRK FAILED same-satellite retry).

Paced on receiver time (`Monitor_Pvt::RX_time`), not wall-clock time -- an offline/replay run processing faster or slower than real time would otherwise cap retries per second of host time, not of the signal's own timeline, making the configured rate mean something different every run depending on host load/speed. Not meaningful before the first fix (`RX_time` isn't ticking yet), so the cooldown simply doesn't engage during that bootstrap phase rather than substituting wall-clock time.

## Configuration

|---|---|---|
| `GNSS-SDR.acquisition_max_retry_rate_hz` | `0` (disabled) | Max acquisition attempts per second for any single PRN+signal. `5` = at most once every 200 ms. |

Defaults off, so an existing `.conf` reproduces today's behavior byte-for-byte unless set explicitly.

## Test plan

- [x] `flowgraph_test` (23 tests, includes `SatelliteVisibilityTest`) passes, from-scratch build off `upstream/next`.
- [x] `clang-format` clean against the CI-pinned version (22).


vladisslav2011:
Hello.
This PR seems to be able to fix a different issue with channels being stuck trying to reacquire the same satellite forever in case of repeated false-positive acquisition (some acquisition implementations are giving too many false-positives). I have not opened an "issue" for it here, but there is a commit with fix attempt and problem description: https://github.com/vladisslav2011/gnss-sdr/commit/82f48bbc2d574ac0764a5013baf36d0c6c2b3b93 (temporary link as this commit id will be updated after a rebase)
Questions:
This PR uses RX_time from the latest PVT as a reference time. What would happen when the PVT got lost after, for example, acquiring 2 or more reflected signals? Isn't it better to rely on a sample counter as a time reference?

joebre:
Good catch, thank you. You're right, and I traced it further: it's actually worse than the PVT just being lost -- \`rtklib_pvt_gs.cc\` only publishes a \`Monitor_Pvt\` status message while \`is_valid_position()\` holds, so \`RX_time\` doesn't reset to invalid on fix loss, it just freezes at its last value indefinitely. Since the cooldown paced off that RX_time, a fix outage would freeze every already-running cooldown timer in place -- a satellite marked right before the outage would never see its elapsed time advance, so it could never leave cooldown for as long as the outage lasted.

Switched to the sample-counter-derived elapsed time (\`channel_status_msg_receiver\`'s \`receiver_time_s\`, published independently of PVT fix validity) that \`SatelliteVisibility::Tick()\` already uses for the same reason -- same sample-counter reference you suggested. That clock is always meaningful (starts at 0.0, monotonic for the life of the flowgraph), so it also drops the old "not meaningful before the first fix" special-casing. Pushed.

I haven't looked at your linked commit for the repeated-false-positive-reacquisition issue yet -- that does sound like a related but distinct problem (this PR throttles retry *rate*, not false-positive *detection*), happy to take a look if useful.

vladisslav2011:
My commit treats failed tracking after successful acquisition exactly like failed acquisition unless a configuration option is set. This does not break anything as the same satellite acquisition will be performed anyways, just a bit later. This PR does the same thing when the satellite acquisition rate limit is exceeded, so it should fix the "stuck channel" issue too.

PR #1104, opened 2026-09-15
Fix two Galileo health-check gaps that silently accept unhealthy satellites

joebre:
## Summary

Two related gaps where a Galileo satellite's broadcast health could be
silently ignored:

- `Galileo_Ephemeris_Store::combined_view()` merged the preferred nav-message
  source's ephemeris straight over the secondary source's entry, discarding
  health/DVS fields the preferred source's message type structurally never
  carries (F/NAV never decodes E1B_HS/E1B_DVS/E5b_HS/E5b_DVS; I/NAV never
  decodes E5a_HS/E5a_DVS). Those fields default to 0 ("OK"), so e.g. an
  E1+E5a configuration preferring F/NAV would present an I/NAV-flagged
  unhealthy satellite as healthy. The merge now fills in only the fields the
  preferred source can't carry, from the secondary source's entry for the
  same PRN.
- `select_galileo_ephemeris()`'s Reduced CED bootstrap fallback had no
  health check at all — `Galileo_Reduced_CED` carries no health/DVS bits of
  its own, so a known-unhealthy satellite with no full ephemeris decoded yet
  would still be accepted through Reduced CED. Now cross-checks the
  almanac's E1B_HS, which every other satellite broadcasts as part of the
  normal almanac rotation.

## Test plan

- Builds cleanly against `next`.
- Reproduced against a live capture: an I/NAV-unhealthy satellite (E1B_HS=1,
  matching independent broadcast-ephemeris references) was previously
  presented as healthy by an E1+E5a (F/NAV-preferred) configuration due to
  the `combined_view()` bug; confirmed fixed.


vladisslav2011:
LGTM.

carlesfernandez:
Galileo_Reduced_CED actually holds E1B/E5b HS and DVS (see [galileo_reduced_ced.h](https://github.com/gnss-sdr/gnss-sdr/blob/next/src/core/system_parameters/galileo_reduced_ced.h)), taken from Word 5. So the change in galileo_ephemeris_store.cc is ok, but the cross-check in rtklib_solver.cc could be detrimental (e.g., it can block the fast-fix path on stale data).

It is ok for you if I just keep the change in galileo_ephemeris_store.cc ?

joebre:
ok

PR #1105, opened 2026-09-15
Add a per-signal broadcast health flag to MonitorPvt's tracked satellites

joebre:
## Summary

`Monitor_Pvt::TrackedSatelliteInfo.used` says whether a tracked
satellite/signal contributed to the fix, but not why it didn't -- below
`PVT.elevation_mask`, RAIM FDE, and broadcast-unhealthy are all
indistinguishable to a monitor client. Adds a `healthy` field (proto
field 8, default `true`) alongside it, in `rtklib_solver.cc`,
`monitor_pvt.h`, `serdes_monitor_pvt.h` and the proto definition.

- GPS: from the decoded ephemeris's `SV_health`, falling back to the
  almanac (broadcast by every satellite) when no ephemeris has been
  decoded for that satellite yet.
- Galileo: resolved per-signal via the existing
  `get_galileo_signal_health()`, with the same almanac fallback when no
  ephemeris is available yet for that signal's service.
- Other systems default to `healthy=true` (unchanged).

## Test plan

- Builds cleanly against `next`.
- Generated protobuf confirms `bool healthy = 8` compiles with matching
  accessors.


vladisslav2011:
LGTM.

PR #1106, opened 2026-09-15
Add ephemeris/almanac-based Doppler prediction as an acquisition search center (fix-only)

joebre:
## Draft: depends on #1097

This is stacked on **#1097** (`Unify pcps_acquisition's Doppler search into a single N-bin API`), which is not yet merged -- this diff will show #1097's changes too until that lands, and the `assist_acquisition_doppler(hz, doppler_num_bins)` call this PR adds relies on `doppler_num_bins`'s "1 == exactly known, search a single bin" semantics from #1097, not the current `doppler_uncertainty` binary scheme on `next`. Marking draft until #1097 merges and this can be rebased onto plain `next`.

## Summary

Today, when acquiring the primary signal of a satellite that `SatelliteVisibility` has already classified visible (elevation known, above the search mask -- `GNSS-SDR.enable_visibility_aware_search`), acquisition still searches the full configured Doppler range, even though ephemeris/almanac plus the receiver's own position/velocity/clock state already predict the Doppler geometrically.

`SatelliteVisibility::PredictedDopplerHz()` computes that prediction: the geometric term from whichever of the satellite's ephemeris/almanac is currently usable (same freshness rules `compute_visible_satellites()` already uses), evaluated at the live PVT fix's position/velocity, plus the receiver's live solved clock drift (`fix_status.user_clk_drift_ppm`) scaled to the signal's carrier.

Wired into `acquisition_manager()`: when dual-frequency assist isn't available, a visible satellite's primary-frequency acquisition now gets a single-bin Doppler search centered on this prediction instead of the full range. No-op (falls back to today's full-range search) unless visibility-aware search is enabled, the satellite is actually visible, and a live fix currently exists -- so behavior is unchanged with the feature off, or before the first fix.

## Scope: fix-only, first step

This PR deliberately does **not** predict Doppler before a fix exists (AGNSS-reference fallback). Before a fix, the geometric prediction would have to assume zero receiver velocity and a static/nominal clock offset, and a single search bin can't absorb the resulting uncertainty -- that needs `GNSS-SDR.clock_frequency_max_error_ppm` / `receiver_max_velocity_m_s` to widen the search bin count accordingly (converting the resulting Hz uncertainty into a `doppler_num_bins` count via the target signal's configured `doppler_step`). That's a meaningfully larger, separately-reviewable change, left for a follow-up PR once this one is in.

## Test plan

- [x] Builds cleanly against a rebase of #1097 onto current `next`.
- [x] `flowgraph_test` (27 tests, including the full `SatelliteVisibilityTest` suite) passes.
- [x] `clang-format` (this branch's CI-pinned version, 22) makes no changes to any touched file.

### DIFF
--- a/src/core/receiver/gnss_flowgraph.cc
+++ b/src/core/receiver/gnss_flowgraph.cc
@@ -1775,12 +1775,12 @@ void GNSSFlowgraph::acquisition_manager(unsigned int who)
                                     // Doppler is exactly known from the already-tracked assisting
                                     // frequency (search_next_signal() returns it already projected
                                     // to this band): restrict the search to a single Doppler bin.
-                                    channels_[current_channel]->assist_acquisition_doppler(estimated_doppler, 0);
+                                    channels_[current_channel]->assist_acquisition_doppler(estimated_doppler, 1);
                                 }
                             else
                                 {
                                     // set Doppler center to 0 Hz and search the full Doppler range
-                                    channels_[current_channel]->assist_acquisition_doppler(0, 1);
+                                    channels_[current_channel]->assist_acquisition_doppler(0, 0);
                                 }
 #if ENABLE_FPGA
                             if (enable_fpga_offloading_)
--- a/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.cc
+++ b/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.cc
@@ -303,9 +303,9 @@ void PcpsAcquisitionAdapter::set_doppler_center(int doppler_center)
 }
 
 
-void PcpsAcquisitionAdapter::set_doppler_uncertainty(unsigned int doppler_uncertainty)
+void PcpsAcquisitionAdapter::set_doppler_num_bins(unsigned int doppler_num_bins)
 {
-    acquisition_->set_doppler_uncertainty(doppler_uncertainty);
+    acquisition_->set_doppler_num_bins(doppler_num_bins);
 }
 
 
--- a/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.h
+++ b/src/algorithms/acquisition/adapters/pcps_acquisition_adapter.h
@@ -103,9 +103,10 @@ public:
     void set_doppler_center(int doppler_center) override;
 
     /*!
-     * \brief Set Doppler uncertainty for the grid search
+     * \brief Set the number of Doppler bins to search, centered on
+     * set_doppler_center() -- see pcps_acquisition::set_doppler_num_bins().
      */
-    void set_doppler_uncertainty(unsigned int doppler_uncertainty) override;
+    void set_doppler_num_bins(unsigned int doppler_num_bins) override;
 
     /*!
      * \brief Returns the maximum peak of grid search
--- a/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
+++ b/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
@@ -131,11 +131,18 @@ pcps_acquisition::pcps_acquisition(const Acq_Conf& conf_)
       d_magnitude_grid_stride(aligned_row_stride<float>(d_effective_fft_size)),
       d_doppler_wipeoffs_stride(aligned_row_stride<gr_complex>(d_fft_size)),
       d_num_doppler_bins(static_cast<uint32_t>(std::ceil(static_cast<double>(2 * d_doppler_max) / static_cast<double>(d_doppler_step)))),
-      d_num_doppler_bins_step1_capacity(std::max(d_num_doppler_bins, 2U)),
       d_num_doppler_bins_step2(conf_.num_doppler_bins_step2),
       d_dump_channel(conf_.dump_channel),
-      d_threshold(conf_.pfa > 0.0 ? compute_threshold(conf_.pfa, d_effective_fft_size, d_num_doppler_bins, conf_.bit_transition_flag ? 1 : conf_.max_dwells) : conf_.threshold),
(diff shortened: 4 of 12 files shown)


src/core/receiver/gnss_flowgraph.cc
```diff
@@ -63,6 +63,7 @@
 #include <iterator>                  // for insert_iterator, inserter
 #include <map>                       // for std::map
 #include <memory>                    // for std::shared_ptr
+#include <optional>                  // for std::optional
```
vladisslav2011: This is  a blocker,
GSL-Lite does not have std::optional implementation....

src/core/receiver/gnss_flowgraph.cc
```diff
@@ -1824,8 +1825,33 @@ void GNSSFlowgraph::acquisition_manager(unsigned int who)
...
-                                    channels_[current_channel]->assist_acquisition_doppler(0, 0);
+                                    // No dual-frequency assist: fall back to a geometric Doppler
+                                    // prediction from ephemeris/almanac (SatelliteVisibility), if
+                                    // this is the primary signal of a satellite already classified
+                                    // visible and a live PVT fix currently exists. No-op (stays
+                                    // nullopt) unless visibility-aware search is enabled --
+                                    // PredictedDopplerHz() itself requires a live fix for now; the
+                                    // AGNSS-reference (no-fix) case is a follow-up.
+                                    std::optional<double> predicted_doppler_hz;
+                                    if (is_primary_freq && satellite_visibility_ && satellite_visibility_->enabled())
```
vladisslav2011: I can't agree with `is_primary_freq` here. L5-only configurations should be supported too. And independent parallel acquisition of L1 and L5/L2 should be supported.

src/core/receiver/satellite_visibility.cc
```diff
@@ -1001,3 +1002,114 @@ bool SatelliteVisibility::IsSearchExcluded(const Gnss_Satellite& sat) const
...
+{
+    if (!enabled_ || !pvt_ptr || fix_status.RX_time < 0.0)
+        {
+            return std::nullopt;
+        }
+
+    // Gnss_Ephemeris::predicted_doppler()/Gnss_Almanac::predicted_doppler()'s
+    // own band codes, keyed by signal the same way SIGNAL_FREQ_MAP is.
+    static const std::map<std::string, int> kBandForSignal = {
+        {"1C", 1}, {"1B", 1}, {"1D", 1}, {"J1", 1}, {"B1", 1},
```
vladisslav2011: I had made a similar mistake: B1C doppler prediction will be completely wrong with this frequency assignment.

src/core/receiver/satellite_visibility.cc
```diff
@@ -1001,3 +1002,114 @@ bool SatelliteVisibility::IsSearchExcluded(const Gnss_Satellite& sat) const
...
+        {"2S", 2}, {"B2", 2},
+        {"L5", 5}, {"5X", 5}, {"J5", 5},
```
vladisslav2011: BeiDou B2a is missing.

src/core/receiver/satellite_visibility.cc
```diff
@@ -1001,3 +1002,114 @@ bool SatelliteVisibility::IsSearchExcluded(const Gnss_Satellite& sat) const
...
+    const double vu_mps = fix_status.vel_u;
+    const gtime_t gps_gtime = gpst2time(static_cast<int>(fix_status.week), fix_status.RX_time);
+    const double rx_time_s = time2gpst(gps_gtime, nullptr);
+    int ref_gps_week = 0;
+    time2gpst(gps_gtime, &ref_gps_week);
+    const std::string system = sat.get_system();
+    const auto prn = static_cast<int>(sat.get_PRN());
+
+    std::optional<double> geometric_doppler_hz;
+    if (system == "GPS")
```
vladisslav2011: QZSS is missing here. It uses GPS almanac/ephemeris structures, so this expression should look like `if (system == "GPS" || system == "QZSS")`

src/core/receiver/satellite_visibility.cc
```diff
@@ -1001,3 +1002,114 @@ bool SatelliteVisibility::IsSearchExcluded(const Gnss_Satellite& sat) const
...
+        {
+            const auto eph_map = pvt_ptr->get_gps_ephemeris();
+            const auto eph_it = eph_map.find(prn);
+            if (eph_it != eph_map.cend() && std::abs(timediff(gps_gtime, eph_to_rtklib(eph_it->second, ref_gps_week).toe)) <= MAXDTOE)
+                {
+                    geometric_doppler_hz = eph_it->second.predicted_doppler(rx_time_s, lat_deg, lon_deg, h_m, ve_mps, vn_mps, vu_mps, band);
+                }
+            else
```
vladisslav2011: GPS CNAV ephemeris may be used to calculate satellite velocity/position/doppler in absence of NAV ephemeris, so it may be better to add another branch with CNAV ephemeris search.

src/core/receiver/satellite_visibility.cc
```diff
@@ -1001,3 +1002,114 @@ bool SatelliteVisibility::IsSearchExcluded(const Gnss_Satellite& sat) const
...
+        }
+    else if (system == "Beidou")
+        {
+            const auto eph_map = pvt_ptr->get_beidou_dnav_ephemeris();
+            const auto eph_it = eph_map.find(prn);
+            if (eph_it != eph_map.cend() && std::abs(timediff(gps_gtime, eph_to_rtklib(eph_it->second).toe)) <= MAXDTOE_BDS)
+                {
+                    geometric_doppler_hz = eph_it->second.predicted_doppler(rx_time_s, lat_deg, lon_deg, h_m, ve_mps, vn_mps, vu_mps, band);
+                }
+            else
```
vladisslav2011: B1C/B2a B-CNAV1 ephemeris may be used to calculate satellite velocity/position/doppler in absence of DNAV ephemeris, so it may be better to add another branch with B-CNAV1 ephemeris search.

PR #1108, opened 2026-09-16
Fix nonblocking acquisition

vladisslav2011:
Implement buffering of all samples, required to perform all non-coherent integrations.
This removes negative effect on performance of `Acquisition_XX.max_dwells` value being greater than 1.

Configuration snippet (only options, that have significant impact on acquisition performance):
```
GNSS-SDR.internal_fs_sps=4000000
Channels_1C.count=15
Channels.in_acquisition=1
Acquisition_1C.implementation=GPS_L1_CA_PCPS_Acquisition
Acquisition_1C.coherent_integration_time_ms=1
Acquisition_1C.pfa=0.015
Acquisition_1C.doppler_max=4400
Acquisition_1C.doppler_step=200
Acquisition_1C.max_dwells=1
;Acquisition_1C.blocking=false
Acquisition_1C.full_grid_search=false
```

Test with `max_dwells`=1
```
...
Current receiver time: 9 min 24 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 23.62 [seconds]
GNSS-SDR program ended.
```

Test with `max_dwells`=20, same machine, same configuration, same dataset
```
...
Current receiver time: 9 min 24 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 24.42 [seconds]
GNSS-SDR program ended.
```

Test with `max_dwells`=20, `blocking`=true, same machine, same configuration, same dataset
```
...
Current receiver time: 9 min 24 s
Stopping GNSS-SDR, please wait!
Total GNSS-SDR run time: 154.46 [seconds]
GNSS-SDR program ended.
```

 - [x] Initial implementation
 - [x] Fix performance regression
 - [x] Optimize state machine transitions
 - [x] Try to request all samples, required to fill the buffer in one big chunk
 - [x] Implement shared buffer pool with number of buffers equal to `Channels.in_acquisition` to reduce memory footprint and improve performance
 - [x] Implement optimization with GNU Radio "history": start non-blocking acquisition as early, as possible and continue buffering samples in background
 - [x] All existing tests passes
 - [x] Test blocking/non-blocking mode with max_dwells set to 1 and >1
 - [x] Test interaction with `bit_transition_flag`
 - [x] Test interaction with `full_grid_search`
 - [x] Test without explicit `Channels.in_acquisition` configuration option
 - [x] Test with fast-producing signal source (file configuration)
 - [x] Test with slow-producing signal source (realtime configuration)
 - [x] Test restarting from standby
 - [x] Test for a deadlock during acquisition thread termination
 - [ ] Long run in realtime configuration


src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
```diff
@@ -165,6 +165,7 @@ pcps_acquisition::pcps_acquisition(const Acq_Conf& conf_)
       d_ifft(gnss_fft_rev_make_unique(d_fft_size)),
       d_grid_doppler_wipeoffs(d_num_doppler_bins_step1_capacity * d_doppler_wipeoffs_stride),
       d_fft_codes(d_fft_size),
+      d_data_buffer(d_acq_parameters.blocking ? d_samples_to_consume : d_samples_to_consume * d_acq_parameters.max_dwells),
```
MathieuFavreau: Since `d_data_buffer` is never used with `d_acq_parameters.blocking`, why allocate any memory for it?

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
```diff
@@ -226,6 +222,16 @@ pcps_acquisition::~pcps_acquisition() noexcept
 }
 
 
+void pcps_acquisition::forecast(int noutput_items __attribute__((unused)), gr_vector_int& ninput_items_required)
+{
+    unsigned ninputs = ninput_items_required.size();
+    for (unsigned i = 0; i < ninputs; i++)
+        {
+            ninput_items_required[i] = d_data_buffer.size();
```
MathieuFavreau: If `d_data_buffer` is not allocated while blocking is true as requested, need to change this to change based on blocking or not

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
```diff
@@ -963,75 +985,72 @@ int pcps_acquisition::general_work(int noutput_items __attribute__((unused)),
...
+                }
+            else
+                {
+                    const auto* in = reinterpret_cast<const gr_complex*>(input_items[0]);  // Get the input samples pointer
+                    std::copy(in, in + samples_to_copy, buffer_ptr + d_buffer_sample_count);
+                }
+
+            d_buffer_sample_count += samples_to_copy;
+            auto n_consume = std::min(static_cast<uint32_t>(ninput_items[0]), d_samples_to_consume);
+            d_sample_count += static_cast<uint64_t>(n_consume);
```
MathieuFavreau, line 1043: Why not `d_sample_count += samples_to_copy`, this seems incorrect in the case where we consume less than `ninput_items[0]` because we are at the end of the buffer

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
vladisslav2011: Thanks.
Good catch. This line was not updated during a rebase as everything seemed to work well.

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
vladisslav2011: This significantly improves the acquisition performance when the number of non-coherent integrations is high. It is possible to consume the whole input buffer, but then we will have to wait for it to fill and delay the start of acquisition. Keeping some samples in the input buffer does not have any impact on performance and does not significantly increase the memory footprint (unlike buffering all samples, required to do all non-coherent integration at once) as all samples reside in a buffer, common to all acquisition and tracking blocks.
The solution to "increased memory footprint" issue is planned (almost ready). I'll push a commit after testing and some refactoring.
\+ It may be unclear why not consume all samples after copying them to a dedicated buffer. This may improve the acquisition performance when a number of channels is small, but the number of `Channels.in_acquisition` is high, for example, it may be equal to the total number of channels. In this case it will be possible to try the next satellite without waiting for the whole buffer to fill.

MathieuFavreau: I am unsure if I don't understand your comment, or you didn't understand mine.

The variable `d_sample_count` is used to tell tracking where acquisition ended (last sample used in acquisition). With your current logic, if let's say we've copied 95% of the samples in the buffer (`d_buffer_sample_count`), `d_samples_to_consume` is 20% of the buffer size (`max_dwells = 5`), and we get samples (`ninput_items[0]`) equals to 10% of the buffer size. In that case, we will copy the remaining samples equals to 5% of the buffer, but `d_sample_count` will increment by samples equal to 10% of the buffer. This will ultimately tell the tracking loop acquisition finished later than it actually did.

MathieuFavreau:
Also, with `max_dwells = 1` and `blocking = false`, you don't need the buffering. I think I would create a member variable `const bool is_buffering` or something like that which is set with `!d_acq_parameters.blocking && d_acq_parameters.max_wells > 1`, and use that everywhere.

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
vladisslav2011: > if let's say we've copied 95% of the samples in the buffer

It is not possible anymore. Forecast always requests all samples, required to fill a buffer, so alignment to a beginning of code phase will be maintained automatically.
In fact, we don't need to give the tracking block exactly the sample, where the acquisition ends. The acquisition block needs any sample, where a code replica starts and performs final alignment itself. There is another reason to avoid giving the tracking block exactly the end of acquisition buffer: BeiDou B3I signal properties. See https://github.com/gnss-sdr/gnss-sdr/issues/663
Other signals seem to be unaffected.

vladisslav2011:
With `blocking = false` the loop in `acquisition_core` will always finish in one iteration, so checking for `blocking = false` should be enough now. In blocking mode it may be better to start non-coherent integrations as early as possible (especially when full grid search is disabled) just to reduce TtFF. Even in case of full grid search it may be better to start non-coherent integration as early, as possible without waiting for all samples to arrive as this should improve acquisition block throughput.

MathieuFavreau:
I didn't mean the behavior is wrong, just that we are doing 1 extra copy of the samples for no reason

vladisslav2011:
In blocking mode samples are copied directly to `d_data_buffer` either directly by `std::copy` or during conversion from `std::complex<int16_t>` by `volk_gnsssdr_16ic_convert_32fc`. The second copying is avoided to preserve the idea of your previous PR.
Am I missing anything?

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
MathieuFavreau: So if I understand right, the goal of your code is with `max_dwell > 1` to offset the acquisition so it doesn't always fall on the same multiple of dwells?

So something like this for `max_dwell = 3`, and assuming a variable `consumed_buffers_while_waiting = 0` in this example, given a buffer is always 3 dwells:
```
Dwell:        0 1 2 3 4 5 6 7 8 9 10
Acquistion 0: <--->
Acquistion 1:   <---> (1 + consumed_buffers_while_waiting * 3)
Acquistion 2:     <---> (2 + consumed_buffers_while_waiting * 3)
```

instead of the old way:
```
Dwell:        0 1 2 3 4 5 6 7 8 9 10
Acquistion 0: <--->
Acquistion 1:       <---> (3 + consumed_buffers_while_waiting * 3)
Acquistion 2:             <---> (6 + consumed_buffers_while_waiting * 3)
```

MathieuFavreau:
What I am saying, is that with blocking **FALSE** and 1 max_dwell, you don't need to buffer the samples, just write them in `d_input_signal`.

Also I replied to a comment you resolved and I can't unresolve

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
vladisslav2011: Exactly.
This significantly improves non-blocking acquisition performance in realtime configurations.

vladisslav2011:
> What I am saying, is that with blocking **FALSE** and 1 max_dwell, you don't need to buffer the samples, just write them in `d_input_signal`.

Thanks. This specific situation is not handled. I'll update the PR for one more time.



> Also I replied to a comment you resolved and I can't unresolve

No problem. The "resolved" status does not disable notifications on new comments.

The CI is failing in completely unrelated place in the PR and not failing in my repository... Maybe a temporary network problem.

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
MathieuFavreau: Would you mind adding a small comment to explain this?

MathieuFavreau:
> The CI is failing in completely unrelated place in the PR and not failing in my repository... Maybe a temporary network problem.

I've seen the same errors on some of my PRs, some tests seem to fail intermittently...

src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc
vladisslav2011: I've added some comments and force-pushed updated commits.
All tests are passing on my machine.
`std::min` is a safeguard as `forecast` is just a forecast and GNU Radio may call `general_work` with different number of samples than we have requested.

vladisslav2011:
This PR has helped me to find a way to reliably trigger a very rare race between `ChannelFsm::stop_acquisition` and `pcps_acquisition::send_positive_acquisition` during GNSS-SDR termination.
Debugging.

Upd:
Nailed it: https://github.com/gnss-sdr/gnss-sdr/issues/1115, https://github.com/gnss-sdr/gnss-sdr/pull/1116

vladisslav2011:
Updated with a concurrent shared buffer pool implementation.
Removed forecast (works well even without it)
Added another rare crash fix. The crash is really hard to reproduce: GNSS-SDR has crashed only once while testing another branch, rebased on top of this branch. The crash dump was analyzed, but the was lost during attempts to reproduce the crash using a clean build of `next` in a clean directory...

A very interesting side effect was observed during testing of this PR: the acquisitions became a lot more deterministic, than before in terms of detected code phase. With this PR almost all detected satellites are showing almost the same code phase in different runs independent of blocking/non-blocking mode, number of channels in acquisition and so on. In `next` every run shows different code phase offset for the same satellite.

carlesfernandez:
I think this is a promising approach, but I still see a couple of problems here:

* Omitting `Channels.in_acquisition` can break nonblocking acquisition. The new pool initialization defaults to zero buffers, while the channel manager defaults to all channels acquiring. With `blocking=false` and `max_dwells`>1, acquisition therefore throws “the pool is empty.” Standalone acquisition blocks also fail without external pool initialization.

* `reserve_buffers()` holds a nonrecursive mutex and calls `foreach()`, which locks it again. Constructing another acquisition block when the pool contains undersized buffers hangs; I reproduced this. `resize_buffers()` has the same issue. Additionally, both callbacks modify buffer copies, so simply fixing the locking would still leave reservation/resizing ineffective.

Behavior change:

- With blocking=false and multiple dwells, acquisition buffers the entire integration window first, then processes it in one worker invocation. Meanwhile, the scheduler can continue consuming input, reducing acquisition-induced backpressure, which is good. Early acceptance remains available, but processing now waits for the whole configured window to arrive, even if the first dwell would succeed. For 20 × 1 ms dwells, that means collecting 20 ms before starting the search.

vladisslav2011:
Thanks for a review. 
In addition to your findings I've found another issue with GNU Radio scheduler during a long run in a realtime input configuration:
```
sched: <block pcps_acquisition (64)> is requesting more input data
  than we can provide.
  ninput_items_required = 50364260
  max_possible_items_available = 80383
  If this is a filter, consider reducing the number of taps.
```
It looks like removing `forecast` was a bad idea.
Switching to `history` + `set_output_multiple` should return old behavior and reduce the delay to first positive acquisition in a configuration with `blocking=false`, `max_dwells`>1 and `full_grid_search=false`. I'll test this approach for one more time and include here if no more issues will be found.
I'll address found issues and update the PR maybe this evening.

vladisslav2011:
Updated.
Configurations missing `Channels.in_acquisition` are fully supported now.
Fixed buffer preallocation and incorrect mutex usage.
Implemented early background thread starting with background buffering by using combination of `set_history` and `set_output_multiple` calls to fix behavior change. Now non-blocking acquisition starts as early as possible, even when it is not possible to buffer all required samples.
A bit more extended testing in different configuration is needed as condition variable usage tend to add deadlocks even in AI-generated code...
Upd:
It crashed during termination. Still not ready for merging...
```
(gdb) bt
#0  0x00007f23b666fe1d in fftwf_execute () from /usr/lib/x86_64-linux-gnu/libfftw3f.so.3
#1  0x000056039b59ad7a in pcps_acquisition::doppler_grid (this=this@entry=0x5603a42fcd90, in=0x7f2398dd5040)
    at /home/vlad/warez/gnss-sdr/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc:683
#2  0x000056039b59cb0f in pcps_acquisition::acquisition_core (this=0x5603a42fcd90, sample_count=479320000)
    at /home/vlad/warez/gnss-sdr/src/algorithms/acquisition/gnuradio_blocks/pcps_acquisition.cc:844
#3  0x00007f23bfdc7bcd in ?? () from /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1
#4  0x00007f23bd7ed6db in start_thread (arg=0x7f201affd700) at pthread_create.c:463
#5  0x00007f23b987461f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

Implementing background buffer filling was a bad idea.

Upd:
This a side effect of adding a new "buffering" state to FSM.

vladisslav2011:
Updated:
The acquisition worker thread termination is performed immediately when `d_active` becomes false now.
This prevents threads from surviving `stop` call and solves "crash during termination" issue for one more time.

Upd:

This PR has processed "forest to city" 50 minute dataset in L1+E1b+B1C+QZSS configuration (non-blocking acquisition, `Channels.in_acquisition=12`) without any freezes/crashes. To my surprise, GNSS-SDR was able to maintain a PVT during whole "deep forest" part, but lost a PVT for ~5 seconds during second forest entry on a highway and things got screwed after driving under a bridge: a PVT was lost for about a minute immediately and got lost multiple times later as satellites became tracked again...

Upd:
Finished 8 hr run in realtime configuration. No crash, no freeze, clean termination.

vladisslav2011:
Another problem was found after a rebase: `pcps_acquisition::wait_if_active()` is not thread-safe and my be called from `pcps_acquisition::stop()` when GNU Radio thread is waiting for the worker thread to join inside of it. In this situation `pcps_acquisition::stop()` will immediately return as `d_worker` id already set to nullptr and the execution will continue to the block destructor leading to a crash during termination...
Protecting `pcps_acquisition::wait_if_active()` with another mutex should solve this problem. Testing.

src/algorithms/libs/buffer_pool.h
```diff
@@ -0,0 +1,181 @@
...
+#include <mutex>
+#include <vector>
+
+/** \addtogroup Algorithms_Library
+ * \{ */
+/** \addtogroup Algorithm_libs algorithms_libs
+ * \{ */
+
+template <class T>
+class BufferPool
```
MathieuFavreau: Personally not a huge fan of static classes, but if we go forward with this, I think it's cleaner to have one static `instance` function that returns the object, and have all the other functions non-static.

In the acquisition class, we can keep the instance reference.

vladisslav2011: Let's try this approach.
1. Rename `get()` to `instance()`
2. Make all other methods non-static

vladisslav2011: Updated (with a separate commit).
If this approach is better, I can squash the commit with 6374ee169bfd2f552b29259bbd9b87a250b61e68 and update the PR.

PR #1112, opened 2026-09-17
Fix scheduler deadlock from hybrid_observables_gs's mandatory clock port

joebre:
## Summary
- `hybrid_observables_gs::forecast()` declares the clock/sample-counter input port mandatory (`ninput_items_required = 1`). GNU Radio will not call `general_work()` at all while that one port is empty, even though the per-channel drain loop a few lines later has no actual dependency on it and already consumes every other port's available input unconditionally.
- This becomes a real deadlock for a signal source that lets several RF bands share one underlying sample buffer via multiple independent readers on a single producer port (GNU Radio's native support for this). If one band's downstream chain temporarily falls behind, it backpressures the shared writer, which stalls every other band's reader too — including whichever one feeds this clock port. With the port mandatory, this block can then never be scheduled again to drain the very channel that's blocking the writer, so nothing ever recovers.
- Fix: relax the requirement to 0, matching every other port, and guard actual epoch computation with a new `d_new_clock_tick` flag (set on a genuine clock push, cleared after use) so it isn't redundantly recomputed on calls that only drain channel backlogs without a fresh clock tick. Since every port can now trivially satisfy its (zero) requirement, add a short sleep on calls that find nothing new anywhere at all, to avoid busy-spinning a CPU core once the scheduler no longer has a natural pacing signal.

## Observed conditions
Reproduced with joint Galileo E1+E5a+E6 processing off a single raw capture at 217 Msps, with each band's own signal conditioner decimating down to a 21.7 Msps working rate (10x). E6's downstream chain (acquisition/tracking/telemetry) runs measurably slower than E1/E5a's at that rate, so it's the band that falls behind and backpressures the shared source buffer, starving the other bands' readers (including the one feeding this block's clock port) and triggering the deadlock described above.

## Test plan
- [x] Builds cleanly (`obs_gr_blocks` target and the full `gnss-sdr` binary)
- [x] `clang-format --dry-run --Werror` clean
- [x] Verified against a 3-RF-band capture at the rates above: reliably froze within the first minute before this fix, ran cleanly for minutes afterward with sane CPU usage


carlesfernandez:
@joebre Can you check if [c285e00](https://github.com/gnss-sdr/gnss-sdr/commit/c285e006a830e6415cf15a2a5e7723b909eac7e1) fixes this issue for you?

joebre:
Hi @carlesfernandez,

Confirmed — `c285e00` fixes it. We built current `next` (includes `c285e00`) and ran it against our original E1+E5a+E6 capture with `enable_e6_tow_map`/`enable_e6_has_rx` both on (the exact combination that previously deadlocked reliably around t=28s): no freeze, E6 HAS pages decoding continuously, and real PVT fixes throughout a run of several minutes.

Thanks for the fix — happy to close #1097/#1106 out of standby whenever you're ready on your end, or let us know if there's anything we can help with there.

joebre:
Hi @carlesfernandez,

We ran a controlled comparison to quantify the tradeoff mentioned earlier: 30 trials each, same config/capture, gnss-sdr-work's original draining approach (unconditional relax + proactive ~1ms poll) vs. current next (c285e00, reactive relax only once a channel already has backlog).

```
gnss-sdr-work:                          next (c285e00):
Trials w/ loss-of-lock: 1/30 (3.3%)     Trials w/ loss-of-lock: 28/30 (93.3%)
Total events: 1                         Total events: 54
                                         E13: 22 (73.3%), E15: 20 (66.7%), E34: 12 (40.0%)
TTFF: 36.0s, 6 SV (both identical)      TTFF: 36.0s, 6 SV (both identical)
```

c285e00 fixes the deadlock correctly, but this is a real, repeatable tracking-stability regression against it. We're still investigating the actual cause and will post again once we've identified it.

joebre:
Quick addendum: the loss-of-lock events are almost entirely on E5a tracking/acquisition specifically, not E1 — worth noting since it points more specifically at the assisted/secondary-band path rather than tracking in general.

PR #1114, opened 2026-09-17
EVK1029: single-port Evk1029Source architecture and file_offset seeking

joebre:
## Summary

Two related, independently-motivated improvements to `Evk1029Source`/`Evk1029SignalSource` (SAPHYRION EVK1029 signal source, from #1091), plus a reference config:

**1. Add `Evk1029SignalSource.file_offset` config** to seek a raw capture to a time of interest instead of always reprocessing from the start. Given in seconds, converted to a byte count using the source's own sampling frequency, then rounded up to the next 64-bit word boundary in `Evk1029Source` before seeking, matching the capture's native packing (16 OBA samples per little-endian 64-bit word). Default 0 (disabled, unchanged behavior).

**2. Give `Evk1029Source` a single output port instead of N coupled ones.** With `role.RF_channels` set to N > 1, it previously declared N separate GNU-Radio-scheduler-coupled `sync_block` output ports and copied the same decoded samples into each of them every `work()` call, so that N RF bands sharing one raw capture would see identical, sample-locked data.

That design has two costs:
- A `gr::sync_block`'s `work()` cannot even be called unless every declared output port currently has space — so one band's downstream chain running temporarily slower than the others stalls production on every port at once, not just its own. If anything downstream depends on one of the *other*, unrelated ports to keep making progress, that coupling can cascade into the whole receiver stalling with nothing left able to run.
- N-1 extra full-buffer `memcpy` calls happen on every `work()` call, at up to the raw capture's sample rate (hundreds of Msps) with large per-call chunks — real, continuous CPU and memory-bandwidth cost for pure duplication.

Fix: `Evk1029Source` now always declares exactly one real output port. Multiple RF bands sharing the capture each get their own `connect()` to that same port instead of separate ports — `gnss_flowgraph.cc`'s existing generic signal-source-to-conditioner wiring already takes this path automatically whenever `output_signature()->max_streams() == 1`, using GNU Radio's native support for several independent readers on one producer port. Each reader then advances at its own pace, and the duplicate-copy cost is gone entirely. This also now matches this project's older `EVK10x9_source`, which used the same single-port-plus-independent-readers design.

`.enable_throttle_control` with `RF_channels > 1` remains unsupported (a single throttle block would itself need the same N-independent-readers treatment) and continues to be ignored with a warning; the warning message is updated to describe why.

**3. Add a reference config** at `conf/File_input/MultiCons/gnss-sdr_EVK1029_Galileo_GPS_E1_E5a.conf` — a complete, ready-to-edit dual-band (Galileo E1+E5a / GPS L1+L5) example demonstrating `EVK1029_Signal_Source` with `RF_channels=2`, the new `file_offset` option, acquisition Doppler-narrowing assist, runtime visibility-aware acquisition search, AGNSS aiding, and the remote monitor/telecommand interfaces. Every key was checked against this branch's source to confirm it's actually implemented here (no options from other, still-unmerged PRs).

## Test plan
- [x] Builds cleanly (full `gnss-sdr` binary, `ENABLE_EVK1029=ON`)
- [x] `clang-format --dry-run --Werror` clean
- [x] Confirmed via log output that the signal-source-to-conditioner wiring correctly takes the single-port/multiple-independent-readers path (`gnss_flowgraph.cc`'s "connecting sig_source_ 0 stream 0 to conditioner N" for every conditioner) instead of the old per-port branch
- [x] Ran against a 3-RF-band capture: acquisition and tracking proceed normally across all bands
- [x] The new reference conf parses and runs cleanly against a real capture with zero config errors/unknown-option warnings


carlesfernandez:
Well, the first finding of the just recently introduced `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON` flag in the CI jobs (see https://github.com/gnss-sdr/gnss-sdr/pull/1110) is here,  thanks @MathieuFavreau  ;-)

The private field `n_streams_` is stored but never read, so it can be safely removed. I'll take care of that, it's pretty simple.

Another detail: `file_offset` have the same meaning and behavior as `seconds_to_skip` in other sources, so I will rename that for consistency.

However, I received a comment from my AI about the configuration example that makes sense:

`InputFilter0.bw` uses a 10 MHz cutoff with a 5 MHz transition before decimating by 10 to 17.75 Msps. firdes puts the cutoff at the centre of the transition band, so the stopband only starts at 12.5 MHz while Nyquist is 8.875 MHz; noise between 8.875 and 12.5 MHz folds onto the 5.25 to 8.875 MHz band edges. That region is inside the E5a/L5 main lobe. Something like bw 7 MHz with tw 3.5 MHz, or less decimation on the E5a chain, avoids it.


I'll merge this PR after the above commented changes, but maybe it is something you can try in your setup, something like:

```
InputFilter0.bw=7000000
InputFilter0.tw=3500000
InputFilter1.bw=7000000
InputFilter1.tw=3500000
```

joebre:
Thanks for the catch, and please pass my thanks to your AI reviewer too -- that's a good one.

Root cause: the `bw`/`tw`/`decimation_factor` on `InputFilter0`/`InputFilter1` in this dual-band conf were carried over as-is from the tri-band (3-band) config, which runs off a different raw sampling frequency and happened to land its aliased noise mostly inside the transition band at that rate. Reusing the same `decimation_factor=10` for the dual-band 177.5 Msps capture didn't reproduce that.

Rather than narrowing `bw`/`tw`, I changed `decimation_factor` from 10 to 7 on both `InputFilter0` and `InputFilter1`: output rate becomes ~25.357 Msps, Nyquist ~12.679 MHz, now just above the 12.5 MHz stopband edge -- so the whole transition band sits below Nyquist and nothing beyond it survives to fold back except already-attenuated stopband content. Updated `GNSS-SDR.internal_fs_sps` to match (25357142.857142857). `bw`/`tw` are unchanged.

Verified in my own setup and updated in `EVK1029_Galileo_GPS_E1_E5a_ref.conf`. Happy to open a follow-up PR with the same `decimation_factor` fix for the merged conf here if useful, just say the word.

PR #1119, opened 2026-09-21
CUDA: fix build on current toolkits and Jetson, add GPU PCPS acquisition engine

phillip-vu:
This PR makes `-DENABLE_CUDA=ON` build and run again on current CUDA toolkits (verified with CUDA 13.2 on JetPack 7.2, and syntax-checked with CUDA 12.4), fixes several bugs in the existing CUDA tracking block, and adds a GPU implementation of the PCPS acquisition search grid that any `*_PCPS_Acquisition` block can enable with one configuration line. Everything was built and tested on an NVIDIA Jetson Orin Nano; a step-by-step Jetson guide with measured results is included.

### 1. CUDA build fixes

The CUDA path had not been buildable with a modern toolkit:

- `tracking/libs/CMakeLists.txt` hardcoded `-gencode arch=compute_30,code=sm_30` (Kepler), which CUDA >= 11 rejects. The flag is removed; `CMAKE_CUDA_ARCHITECTURES` is now honored if set by the user, detected from the device tree on NVIDIA Jetson (Orin: 87, Xavier: 72, TX2: 62, Nano: 53), or set to `native` with CMake >= 3.24.
- `CMAKE_CUDA_STANDARD` was pinned to 14 while the project compiles C++17/20; it now follows the host standard.
- `find_package(CUDAToolkit)` is used and `CUDA::cudart` / `CUDA::cufft` are linked explicitly (the previous code relied on implicit linking through the CUDA language, which is not enough for the new cuFFT dependency).
- `CUDA_GPU_ACCEL=1` is defined project-wide (via `add_compile_definitions`) instead of per-target, because `pcps_acquisition.h` now changes its class layout under that macro and every translation unit including it must agree.
- `-Wno-psabi` is restricted to the CXX language instead of being dropped whenever CUDA is enabled (nvcc rejects it).
- `cuda_multicorrelator.cu`: `cudaDeviceProp::deviceOverlap` and two cuFFT enumerators were removed in CUDA 13.

### 2. Bug fixes in `GPS_L1_CA_DLL_PLL_Tracking_GPU`

- **Cross-block data race.** The carrier wipe-off and the multi-tap correlator ran in one kernel, separated only by `__syncthreads()`. That only orders threads within a block, but every block reads the whole wiped vector. They are now two launches on the same stream.
- `cudaHostAlloc(..., cudaHostAllocMapped || cudaHostAllocWriteCombined)`: the logical OR evaluates to `1` == `cudaHostAllocPortable`, so the buffers were never requested as mapped. Now `cudaHostAllocMapped`.
- `cudaDeviceReset()` was called from the per-channel `free_cuda()`, tearing down the context under every other channel still tracking. Removed (the process-wide reset in `main()` stays).
- `cudaFree()` was called on device aliases of host-mapped buffers that the caller releases with `cudaFreeHost()`.
- `GPU_Complex` operators made const-correct; stream/device members initialized.

### 3. GPU acquisition grid (`CudaPcpsEngine`)

`src/algorithms/acquisition/libs/cuda_pcps_engine.{h,cu}` evaluates, for all Doppler bins at once, `|IFFT(FFT(in .* w_k) .* conj(FFT(code)))|^2` with a batched cuFFT plan and three small kernels (wipe-off, code multiply, magnitude/non-coherent accumulation). One stream and one plan per acquisition block, so channels acquiring concurrently overlap on the device. The header contains no CUDA types.

`pcps_acquisition::doppler_grid()` dispatches to the engine when the block is configured with

```
Acquisition_1C.use_cuda=true          ; per block, or
GNSS-SDR.use_cuda_acquisition=true    ; for every PCPS block
```

and otherwise runs the unchanged CPU loop (now `doppler_grid_cpu()`). Peak search, CFAR statistics, `make_two_steps`, Doppler narrowing, `full_grid_search`, dumping and the monitor output are untouched and run on the host as before, so results are identical to the CPU path. If the device cannot be initialized (or the build has no CUDA), the block logs a warning and continues on the CPU, so configurations stay portable. The wipe-off grid and `conj(FFT(code))` are uploaded only when they change; the cuFFT plan is built and warmed up at upload time so that the first dwell has deterministic latency in non-blocking mode.

### 4. Tests and benchmark

- `CudaPcpsEngineTest` (in `run_tests` when `ENABLE_CUDA=ON`): compares the GPU grid against a CPU reference (volk + `gr::fft`) sample by sample for a single dwell, 4-dwell non-coherent accumulation, `bit_transition_flag` mode and 4 ms coherent integration; checks invalid-usage reporting; prints a CPU/GPU timing.
- `GpsL1CaPcpsAcquisitionCudaTest`: runs the full `GPS_L1_CA_PCPS_Acquisition` adapter on the existing 4 Msps captures with `use_cuda=true`, checks the expected Doppler/delay, and checks that the CPU and CUDA paths return the same estimate with and without `make_two_steps`.
- `benchmark_pcps_grid` (with `ENABLE_BENCHMARKS=ON`): Google Benchmark sweep of the CPU baseline and the GPU engine over FFT sizes 2000 to 40000 and 21, 41 and 81 Doppler bins, with `dwells/s` and `grid_cells/s` counters. The CPU part builds without CUDA too.

### 5. Documentation

- `docs/JETSON.md`: prerequisites, build, architecture selection, verification, benchmark, measured results, receiver usage, and troubleshooting on NVIDIA Jetson.
- `utils/scripts/jetson-build.sh`: scripted version of the above.
- README (CUDA section and `Acquisition_1C.use_cuda` parameter), changelog, and `conf/Other/gnss-sdr_GPS_L1_gr_complex_gpu.conf` updated.

### Results (Jetson Orin Nano Super, JetPack 7.2 / CUDA 13.2, 25 W mode)

Unit tests: GPU grid matches the CPU grid to a relative error below 5e-7 (single-precision rounding); same peak bin/index; same acquisition estimates as the CPU path on the real captures (1700 Hz / 523 samples; two-step 1740 Hz / 911 samples).

`benchmark_pcps_grid`, one complete search grid, wall-clock including host to device and device to host transfers (CPU = one core, as `pcps_acquisition` uses it):

| `fft_size` | bins | CPU (us) | GPU (us) | speed-up |
| ---------: | ---: | -------: | -------: | -------: |
|       2000 |   21 |      759 |      128 |     5.9x |
|       4000 |   41 |     3085 |      336 |     9.2x |
|       8000 |   81 |    13600 |     1180 |    11.5x |
|      16000 |   41 |    15091 |     1334 |    11.3x |
|      20000 |   81 |    46426 |     3097 |    15.0x |
|      40000 |   81 |   101573 |     8023 |    12.7x |

(Full 18-point table in `docs/JETSON.md`.)

Receiver end-to-end on the public `2013_04_04_GNSS_SIGNAL_at_CTTC_SPAIN` capture (60 s, 8 channels, `Channels.in_acquisition=4`, 3 runs each): CPU and CUDA acquisition give the same fix (41.2748 N, 1.9877 E), the same time to first fix, and the same 69 PVT solutions. Processing 60 s of signal takes 12.3 to 12.7 s on the CPU path and 10.1 to 10.7 s with CUDA acquisition.

### Notes for reviewers

- All existing CPU acquisition tests (`GpsL1CaPcpsAcquisitionTest`, `GalileoE1PcpsAmbiguousAcquisitionTest`, `GpsL2MPcpsAcquisitionTest`, `GlonassL1CaPcpsAcquisitionTest`, `GalileoE5bPcpsAcquisitionTest`, `BeidouB1iPcpsAcquisitionTest`) still pass on the same build.
- Without `ENABLE_CUDA`, the only behavioral change is a warning if `use_cuda=true` is set in a configuration file.
- The GPU tracking block is still marked experimental; this PR only makes it build and fixes the bugs above, it does not change its algorithm.
- Not yet characterized: aggregate GPU throughput with many channels acquiring simultaneously (each channel has its own stream; the benchmark measures a single engine).


carlesfernandez:
Thanks @phillip-vu for this awesome contribution! 

please remember that you need to sign your commits, see instructions [here](https://gnss-sdr.org/docs/tutorials/using-git/#how-to-work-with-the-source-code) before they can be accepted into the `next` branch.

For existing commits, you can do (from your branch, and assuming that your remote `upstream` points to `https://github.com/gnss-sdr/gnss-sdr`):

```
git fetch upstream
git rebase --signoff upstream/next
git push --force-with-lease
```

phillip-vu:
Closing in favor of https://github.com/gnss-sdr/gnss-sdr/pull/1120, which carries the same change set with signed commits under my main account. Sorry for the noise.

